PR #11, opened 2026-08-08
fix(db): stop `db rebuild` from destroying the committed intent files

reinaldosaraiva:
`hedgehog db rebuild` destroys the committed intent files it is supposed to replay. On the project where this first surfaced, one module lost 10 requirements and another lost 22 in a single invocation. Because the rebuild fires **automatically** when the DB is missing but `.hedgehog/intents/` exists, a user can lose their planning record without ever typing the command.

Both causes are confirmed in the current code, with a reproduction for each.

## Cause 1 — replay ignores `depends_on`

`src/db/rebuild.mjs` sorts intents alphabetically and replays in that order:

```js
return entries.filter((name) => name.endsWith('.json')).sort();
```

`intent_dependencies.depends_on_intent_id` is `REFERENCES intents(id)` (`schema.mjs:27`) and `PRAGMA foreign_keys = ON` is set in both `schema.mjs` and `init.mjs`, so replaying an intent before its dependency throws `FOREIGN KEY constraint failed`.

## Cause 2 — write-before-insert, plus a normalize asymmetry

`src/db/intent.mjs` builds `requirements` from the *input* fields while the returned record carries the *output* field:

```js
const requirements = [];
for (const [kind, field] of Object.entries(REQUIREMENT_KIND_FIELDS)) {
  const statements = record[field] ?? [];   // rules / constraints / acceptance
```

…and writes the file before the insert:

```js
await writeFile(`${INTENTS_DIR}/${intent.id}.json`, JSON.stringify(intent, null, 2));
...
db.exec('BEGIN IMMEDIATE');
```

So re-normalizing an already-written file yields zero requirements, and the file is rewritten before — and regardless of — the insert.

## Two details that make it worse

- **The causes compound.** The FK crash in cause 1 happens *after* the current file has already been rewritten, so a rebuild that throws still leaves emptied records behind. That is why the original loss was total on a single attempt.
- **`rebuildDb`'s `intentsDir` option was never honoured on the write side** — `addIntent` hard-coded `INTENTS_DIR`, so a rebuild pointed at one directory could rewrite another. Removing writes from the replay path retires this.

## The fix

**`fix(db): stop intent writes from emptying the committed record`**
- `normalizeIntent` now reads `requirements` alongside the input arrays (preserving ids, deduplicating), so `normalize(normalize(x)) === normalize(x)`.
- `addIntent` inserts first, writes second.
- New `assertNonDestructiveWrite` refuses to overwrite an intent file holding N > 0 requirements with one holding none, and refuses to overwrite a file it cannot parse. It runs before the insert, so a refusal leaves nothing half-done.
- The DB insert is split out as `insertIntentRows` — a file-free path.

**`fix(db): replay intents in dependency order and never rewrite them`**
- Topological sort over `depends_on`, tie-broken by filename; the whole order resolves before any row is written.
- Cycles raise `intent depends_on cycle detected: alpha -> gamma -> beta -> alpha`.
- A `depends_on` naming neither a file nor an existing row is reported explicitly instead of surfacing as an FK error.
- Replay calls `insertIntentRows`, so **the rebuild path never writes an intent file at all**. A rebuild reads the permanent record; it does not rewrite it.

## Reproduction

`node repro/run.mjs` — 4 reproductions, `node:sqlite`, plain assertions, each in its own `mkdtemp` (no test framework is available in this repo). Fixtures use `automation → card → list → board`, where alphabetical order contradicts dependency order, with `card` carrying 22 requirements and `list` 10 — the real loss, reproduced.

Before, all 4 fail:

```
FAIL  2. every intent JSON file is byte-identical after the rebuild
      c-card.json: sha 229f65636910 -> 377815014018, 2715B -> 161B,
        requirements 22 -> 0  <-- PERMANENT RECORD EMPTIED
FAIL  1. rebuild completes without throwing
      rebuild threw: FOREIGN KEY constraint failed
```

After, all 4 pass, with sha256 per file plus `git status --porcelain .hedgehog/intents` as an independent witness. Also verified end-to-end through the real CLI in a throwaway repo, including the automatic rebuild that `ensureDb` triggers when the DB is missing: sha256 of all three intent files unchanged, `git status` clean. `node scripts/check.mjs` passes.

Commit 1 was verified green on its own with `rebuild.mjs` reverted, which isolates cause 2 as fully addressed by the `intent.mjs` change alone.

### DIFF
--- a/repro/add-intent-safety.mjs
+++ b/repro/add-intent-safety.mjs
@@ -1,0 +1,143 @@
+// Reproduction: the two properties of the write path that made the
+// rebuild destructive in the first place, plus the safety net.
+//
+//   A. `normalizeIntent` must round-trip its OWN output. It writes
+//      `requirements` but reads `rules`/`constraints`/`acceptance`, so
+//      feeding a written intent file back through it yields zero
+//      requirements. That asymmetry is what emptied the record.
+//
+//   B. `addIntent` must not write the intent file before the DB insert.
+//      With write-first, an insert that fails leaves a rewritten (and
+//      possibly emptied) permanent record behind with nothing in the DB
+//      to show for it.
+//
+//   C. Overwriting a non-empty intent file with a zero-requirement one
+//      must fail loudly rather than silently empty a committed record.
+//
+// Runs entirely inside a temp dir created by the harness.
+
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { normalizeIntent, addIntent } from '../src/db/intent.mjs';
+import { openDb, dbInit, DB_PATH } from '../src/db/init.mjs';
+import {
+  intentFixture,
+  makeProject,
+  snapshotIntents,
+  diffSnapshots,
+  check,
+  assert,
+  assertEqual,
+  finish,
+} from './harness.mjs';
+
+console.log('repro: addIntent / normalizeIntent must not destroy a written record\n');
+
+// 14 rules + 4 constraints + 4 acceptance = 22 requirements, the count
+// the real project's `card` module lost.
+const seed = intentFixture({ id: 'card', rules: 14, constraints: 4, acceptance: 4 });
+const cardPath = () => join(root, '.hedgehog', 'intents', 'card.json');
+
+const root = await makeProject([seed]);
+console.log(`  temp project: ${root}`);
+console.log(`  card.json carries ${seed.requirements.length} requirements\n`);
+
+await dbInit(DB_PATH);
+const db = openDb();
+
+// --- A. normalizeIntent round-trips its own output --------------------
+
+check('A. normalizeIntent round-trips its own output', () => {
+  const fromInput = normalizeIntent({
+    id: 'roundtrip',
+    goal: 'g',
+    outcome: 'o',
+    rules: ['r1', 'r2'],
+    constraints: ['c1'],
+    acceptance: ['a1'],
+  });
+  assertEqual(fromInput.requirements.length, 4, 'requirements from the input shape');
+
+  const again = normalizeIntent(fromInput);
+  assertEqual(again.requirements.length, 4, 'requirements after re-normalizing the output shape');
+  assertEqual(
+    JSON.stringify(again.requirements),
+    JSON.stringify(fromInput.requirements),
+    'requirements should survive a round trip unchanged',
+  );
+});
+
+check('A2. normalizeIntent preserves a real on-disk intent record', () => {
+  assertEqual(
+    normalizeIntent(JSON.parse(readFileSync(cardPath(), 'utf8'))).requirements.length,
+    seed.requirements.length,
+    'requirements after normalizing the committed file',
+  );
+});
+
+// --- B. a failed DB insert must not have touched the file -------------
+
+const beforeAdd = await snapshotIntents(root);
+await addIntent(db, seed);
+const afterAdd = await snapshotIntents(root);
+
+check('B1. a successful add leaves the committed record byte-identical', () => {
+  const problems = diffSnapshots(beforeAdd, afterAdd);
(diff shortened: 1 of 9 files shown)


skyf0xx:
Confirmed, recommended first in merge order — smallest blast radius, adds a guard rather than relaxing one. `rebuild.mjs:28`'"'"'s alphabetical sort and the missing `UPDATE tasks` path are both real.

Needs rebase onto current master (mechanical — your files are untouched by 7450c84/a300678).

---

reinaldosaraiva:
Update: an independent automated review of this branch raised a window my reorder opened, and it reproduces — worse than reported.

**Database commit strands intent state.** Insert-first/write-second fixed the original bug, but traded it for the mirror image: a failed `writeIntentFile` after a successful insert leaves the intent in the derived DB only.

Reproduced by `chmod 0555` on `.hedgehog/intents`, driven through the real CLI:

| | before |
|---|---|
| `intent add` | `Failed to add intent: EACCES`, exit 1 |
| DB `intents` | `[{"id":"board","status":"proposed"}]` — stranded |
| disk | empty |
| retry once writable | `UNIQUE constraint failed: intents.id`, exit 1 |
| rebuild after deleting the DB | `No build graph found` — **`board` gone permanently** |

That last row is the part worth stressing: the DB row was the *only* copy, and the DB is gitignored and routinely thrown away. The retry constraint made it unrecoverable through any supported command.

**Fixed by rollback rather than idempotent retry**, implemented by moving the file write *inside* the transaction, immediately before `COMMIT`:

```js
db.exec('BEGIN IMMEDIATE');
try {
  runIntentInserts(db, intent);              // rejected records never reach the file
  await writeIntentFile(intent, intentsDir); // a failed write rolls the inserts back
  db.exec('COMMIT');
} catch (err) { /* ROLLBACK */ throw err; }
```

The invariant worth holding is that **the DB never holds an intent the record doesn't**. An upsert-style retry would keep the strand and paper over it, and would make `intent add` silently overwrite an existing intent — the same class of quiet overwrite as the original bug, and directly at odds with `assertNonDestructiveWrite`.

Moving the write up next to `assertNonDestructiveWrite` (before the insert) does not work: that is the original bug's order and re-breaks the rejected-record case. Between the inserts and the `COMMIT` gets both properties, because every constraint in play is immediate — so "the inserts returned" already means "the DB accepted this intent" before anything touches disk.

Residual, and it is the recoverable direction: `COMMIT` itself failing after the file landed leaves a file without a row — the file is the source of truth, `db rebuild` replays it, and re-running the add passes the guard.

`writeIntentFile` now also writes a sibling temp file and renames it into place, because a crash mid-write would otherwise leave a truncated record that `assertNonDestructiveWrite` then refuses to overwrite — a partial write becoming its own dead end.

New reproduction: 4 of 7 assertions fail before, all 7 pass after. Existing five still pass, including the per-file sha256 and the `git status --porcelain` witness. `scripts/check.mjs` green.

skyf0xx:

---

skyf0xx:
Landed on master at 71ca891, rebased onto current master (past 7450c84) since this branch predated that fix. Merge conflicts in src/db/intent.mjs and src/db/rebuild.mjs — both import-only collisions with #29's once-layer additions (CORE_INTENT_ID, CORE_MODULE) — resolved to keep both sides' logic; verified `markCompletedTasks`'/`loadCommitSubjects`'s Map-based once-layer reconciliation from #29 still lines up with this PR's dependency-ordered replayIntents.

Verified before landing: `node scripts/check.mjs` passes, and all of this PR's own reproductions (add-intent-safety, add-intent-write-failure, rebuild-cycle, rebuild-out-of-order, rebuild-preserves-intent-files) pass against the merged code. repro/05-shipped-cores-unchanged.mjs fails, but that's pre-existing on master independent of this PR (stale verify_command fixture) — tracked separately, not caused by this change.

---

Issue #9, opened 2026-08-08
[bug] Nothing pushes a core's verify commands to assert observable effect, so gates pass on reported state: five false greens and one false red in one build


reinaldosaraiva:
From the same 36-task authored-core build as #8 (6 modules × 6 layers, backend
+ k3s, GitOps via ArgoCD). Where #8 is about scope, this is about what a
`verify` command is allowed to assert.

## The pattern

`hedgehog-core-design` accepted every command below without objection, and
`hedgehog verify` ran each of them exactly as written. All of them ask the
system **what it reports about itself** rather than **what it actually does**.
That distinction is the whole difference between a gate and a formality.

## Five green gates over false statements

| What the verify asked | What was true |
|---|---|
| `grep -q Healthy` on an ArgoCD Application | it managed **zero** resources — an empty set is trivially healthy |
| `kubectl rollout status` on both Deployments | the ingress was serving **a different project's** app from another namespace |
| `kubectl wait --for=condition=complete job/db-migrate` | the Job was `Complete` from a *previous module*; a Kubernetes Job template is immutable, so `kubectl apply` never re-ran it and the new migration never reached Postgres |
| `kubectl apply` + `rollout status` with `:latest` images | the running pods were an hour old and **did not contain the new code** |
| six green `deploy` layers, ArgoCD `Synced`/`Healthy` at the right SHA | the last module's primary API route answered **404**, served by the wrong app (filed separately) |

## And once in the opposite direction

The same class of question also produced a **false red**. After a deploy where
everything was genuinely correct in the cluster — migrations 4 of 4, new images
in the pods, five HTTP assertions all returning what was expected — `hedgehog
verify` failed the task. The only failing assertion was ArgoCD's
`status.health.status`, which read `Progressing` for **more than six minutes**
over pods that were `1/1 Running` and a rollout long finished. Annotating the
Application with `argocd.argoproj.io/refresh=hard` flipped it to `Healthy` in
under eight seconds.

The gate was not observing the cluster. It was reading a summary the controller
had not recalculated yet — and the verify runs that query immediately after
mutating the resources, which is precisely when the cache is guaranteed stale.

This matters because it settles the principle in both directions: **reported
state does not work as a gate, neither to approve nor to reject.**

## Aggravating factor

The `deploy` layer's verify command is byte-for-byte identical across all six
modules, because it contains no `{module}` token — unlike the schema, service,
api and ui layers, whose verify commands anchor a test filter to the module's
[...]

PR #12, opened 2026-08-08
feat(cli): add show, retry and targeted claim, so recovery stops needing raw SQL

reinaldosaraiva:
Across one full 36-task authored-core build, the operator had to drop out of the CLI and run raw `UPDATE` statements against `.hedgehog/hedgehog.db` **eight separate times** — for situations that are entirely normal in the intended workflow. This adds the missing commands, with reproductions.

## The four gaps, confirmed in the current code

**1. A failed verify is a dead end.** `verify.mjs` blocks the task and clears the lease. The recovery the loop skills document — "fix it and re-run `hedgehog verify <task-id> --owner <owner>`" — is **impossible**: `claimForVerify` asserts `if (task.status !== 'building' || task.lease_owner !== owner) throw`, and `releaseTask`'s UPDATE is `WHERE id = ? AND lease_owner = ? AND status = 'building'`. Nothing in the CLI writes `blocked → anything`. A failed verify is the *expected* case of the loop — build, gate rejects, fix, retry — not an anomaly.

**2. A scope violation is identical.** Same `setTaskStatus(..., 'blocked', ...)` path, same three refusals.

**3. `claim` printed no packet, and the packet then became unreachable.** The documented flow says claim returns a packet with STATUS / ALLOWED SCOPE / VERIFICATION per task; `claimCommand` printed only the id and lease expiry. And because `next.mjs`'s `READY_TASK_SQL` requires `status IN ('planned','ready') AND t.lease_owner IS NULL` while `claimOne` sets `status = 'building'`, `hedgehog next` then reports "No ready task" — so no command could print the scope and verify command the operator is supposed to delegate. Reading the `tasks` row out of SQLite by hand was the only path.

**4. No targeted claim.** `claimTasks` walks `ORDER BY t.priority, t.id` greedily and skips any candidate where `conflicts(...) !== null`. Since `conflicts` returns `'exclusive'` if *either* side is exclusive, an exclusive task is skipped whenever any non-exclusive candidate sorts ahead of it — **permanently**, not just while work is in flight. As the first layer of each module, a starved `exclusive` infra layer stops that whole module from starting.

## What this adds

- **`hedgehog show <task-id>`** — the `next` packet for any task at any status. `formatNext` still passes `'READY'`, so `next` output is byte-identical (verified by diffing `next`/`status`/`ready` output before vs after).
- **`hedgehog retry <task-id>`** — `blocked → planned` for every `blocked_reason`, clearing `blocked_reason` and all three lease columns in one UPDATE. The schema's `CHECK ((lease_owner IS NULL) = (status NOT IN ('building','verifying')))` holds.
- **`hedgehog claim <task-id> --owner <o>`** — keeps every fan-out precondition (status, lease, dependencies, conflict with in-flight work) and reports *which* one refused, exiting non-zero. Deliberately **no** `--force`: releasing or finishing the neighbour is the fix, not overriding the conflict rule.
- **`claim` now prints the full packet** for each task it hands out, after the unchanged summary lines.
- **Every write command prints `db  <absolute path>`**, and `retry` / targeted `claim` / `release` / `renew` / `show` exit non-zero with "No such task" on a zero-row match.

That last point closes a failure mode observed directly: a hand-written recovery script used a relative DB path, was run from the wrong directory, opened a different database, matched zero rows, printed `undefined`, and exited **zero**. A silent green success over the wrong database.

- **The three loop skills and the `status` hint now point at `retry`** instead of the re-verify that cannot work.

## Reproductions

`node repro/run-all.mjs` — standalone Node, `node:sqlite`, plain assertions, a throwaway git + graph project per case (no test framework is available in this repo).

- **Before** (implementation stashed): 5 of 5 scripts fail — `Unknown command: retry`, claim printing no `ALLOWED SCOPE`, the targeted claim not existing, no db path in output.
- **After**: all 8 named cases pass, including "a recovery command run in the wrong directory refuses".

`node scripts/check.mjs` is green, and `bin/cli.mjs` parses and runs at every one of the commits.

## Honest caveats

- Batch `claim` returning zero still exits **0**. That is a documented state the loop skills depend on ("if claim returns nothing and the graph isn't done…"), so changing it would break existing behaviour. Only the targeted commands fail loudly on no match.
- The `bin/cli.mjs` hunks were split into an observability commit and a new-commands commit; the import line lands in the earlier one, so that commit imports two symbols it does not yet use. Cosmetic — the CLI runs at every commit — and left alone rather than risking a rebase.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -334,7 +334,10 @@ ${bold('Usage')}
   npx @skyf0xx/hedgehog intent add [flags]        add an intent (rules/requirements/dependencies)
   npx @skyf0xx/hedgehog intent add --file <path>  add an intent from a JSON file
   npx @skyf0xx/hedgehog next                      print the task packet for one ready task
+  npx @skyf0xx/hedgehog show <task-id>            print the task packet for any task, at any status
   npx @skyf0xx/hedgehog claim --owner <owner> [--count <n>]   atomically claim up to n ready tasks
+  npx @skyf0xx/hedgehog claim <task-id> --owner <owner>   claim one specific task (breaks a starvation tie)
+  npx @skyf0xx/hedgehog retry <task-id>           return a blocked task to planned, so it can be rebuilt
   npx @skyf0xx/hedgehog release <task-id> --owner <owner>   hand a claimed task back to ready
   npx @skyf0xx/hedgehog renew <task-id> --owner <owner> [--minutes <n>]   extend a held lease
   npx @skyf0xx/hedgehog verify <task-id> --owner <owner>   run scope + verify checks, commit on pass
@@ -785,8 +788,11 @@ async function nextCommand() {
         const reason = BLOCKED_REASON_LABELS[task.blocked_reason] ?? task.blocked_reason;
         console.error(`  ${red('✗')} ${bold(task.id)}   ${task.layer}   ${dim(reason)}`);
       }
+      // `verify` refuses anything that isn't `building` and leased to
+      // the caller, so a blocked task has to go back through the queue
+      // — retry, claim, then verify — rather than straight to verify.
       console.error(
-        `\nFix the work, then re-run ${bold('hedgehog verify <task-id>')}.\n`,
+        `\nFix the work, then ${bold('hedgehog retry <task-id>')} and claim it again.\n`,
       );
       process.exitCode = 1;
       return;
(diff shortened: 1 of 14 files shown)


reinaldosaraiva:
Heads-up on an overlap: #19 also makes `hedgehog claim` emit the full packet. The two were written independently against master and will conflict textually on that one commit.

The rest of #19 (intent GOAL/OUTCOME labels, an intent-close restatement, and a debt table) is independent of this PR. Whichever of the two you prefer to land first, I'll rebase the other to drop the duplicated claim-packet change.

skyf0xx:
Confirmed real and clean — `claim.mjs:119`/`verify.mjs:294` confirm a `blocked` task is genuinely unreachable today. `retry` re-enters the normal readiness path, doesn't skip verify.

Correction: I said elsewhere this addresses #9 — it doesn'"'"'t, different defect (recovery vs. verify-asserts-reported-state). Noted there.

Overlaps #19 (packet-claim, you'"'"'ve offered to drop it) and #29 (exclusive-starvation, opposite approach) — worth deciding ownership before merge.

---

skyf0xx:

---

skyf0xx:
Landed on master at b67f008, rebased onto current master (past 7450c84 and #29's layer-cardinality work), since this branch predated both.

Two real conflicts, both resolved to keep both sides:
- `src/db/next.mjs` — #29 added once-layer-aware wording to `formatPacket`'s WHY NOW section; this PR added `incompleteDeps` listing to the same section for `show`'s use case. Kept both, as two independent lines rather than competing changes.
- `repro/run-all.mjs` — both this PR and #29 independently wrote a runner script. Kept #29's version (spawnSync with `--experimental-sqlite` passed explicitly), since `node:sqlite` on this Node version requires that flag and this PR's version omitted it.

One real behavioral interaction, not just a merge conflict: `repro/04-claim-specific-task.mjs`'s scenario was built entirely around exclusive-task starvation — the fan-out never picking `ALPHA-ZINTEGRATE` because it sorted after non-exclusive candidates. #29 fixed exactly that starvation (`ORDER BY t.priority, t.exclusive DESC, t.id`), so the premise no longer holds: the batch claim now correctly claims the exclusive task first. That's the intended improvement, not a regression. Rewrote the reproduction to exercise what targeted claim still adds post-#29 — claiming one task without claiming the rest of the ready set, and refusing when the requested task conflicts with work already in flight — rather than leave a test asserting stale behavior. `show`, `retry`, and the DB-path/exit-code fixes were all unaffected and verified as-is.

Verified before landing: `node scripts/check.mjs` passes, and all 5 of this PR's reproductions pass, plus the full 14-script combined suite (only the pre-existing, unrelated `05-shipped-cores-unchanged.mjs` fixture-staleness failure remains — not caused by this PR).

---

PR #13, opened 2026-08-08
fix(core): unescape quoted scalars in the core.yaml parser

reinaldosaraiva:
The hand-rolled `core.yaml` parser does not unescape quoted scalars. It recognises a quoted scalar by its first and last character and strips exactly those two, with no unescaping step at all — `src/db/core.mjs`, `parseScalar` before this change:

```js
function parseScalar(raw) {
  const s = raw.trim();
  if (
    (s.startsWith('"') && s.endsWith('"')) ||
    (s.startsWith("'") && s.endsWith("'"))
  ) {
    return s.slice(1, -1);
  }
  return s;
}
```

## Why it matters

An escaped quote is the only single-line way to write a nested quote inside a shell `verify` command. Parsing

```yaml
verify: "export PATH=\"$HOME/bin:$PATH\"; echo PATH=$PATH"
```

and then running **what the parser returned** through `sh` gives:

```
parsed:  export PATH=\"$HOME/bin:$PATH\"; echo PATH=$PATH
sh says: PATH="/home/u/bin:/usr/local/sbin:..."
```

The parasitic quotes end up inside the value and the resulting `PATH` is invalid. The failure is silent in the worst way: the YAML loads, `loadCore` validates, every layer appears correctly, and only the runtime behaviour is wrong. It was found only by extracting the string exactly as the parser returned it and running *that*, rather than the string as written in the file.

## Two more defects with the same root cause

Both found while confirming the first:

- **`''` in a single-quoted scalar is not unescaped either.** `'echo it''s fine'` returns `echo it''s fine`.
- **`stripComment` toggles quote state without honouring backslashes**, so `verify: "echo \"tag#1\" && test -f a"` parsed to `"echo \"tag` — silently, since the mangled remainder no longer ends in a quote and falls through the `startsWith`/`endsWith` test unstripped.

## The fix

Implementing the escapes rather than rejecting them: the single-quoted fallback was broken the same way, so rejection would have left core authors with no working form for a nested quote. The unescaping is ~40 lines with no new dependency.

- Double-quoted scalars take the YAML 5.7 escape set: `\" \\ \/ \n \t \r \0 \a \b \v \f \e \<space> \N \_ \L \P`, plus `\xNN`, `\uNNNN`, `\UNNNNNNNN`.
- Single-quoted scalars take `''` as one literal quote, with no backslash escapes.
- The closing quote is found by scanning with escapes honoured, not by testing the last character.
- `stripComment` skips the character after a backslash while inside a double-quoted scalar.
- Where the parser still cannot represent something — unknown escape, trailing backslash, unterminated or improperly closed quote — it **throws**. The residual gap is loud, never a plausible-looking wrong string.
- The header comment documents the supported set and what remains outside it.

## Evidence

`repro/yaml-scalar-escapes.mjs` — plain assertions, prints expected vs actual, non-zero exit (no test framework is available in this repo). It includes the method that found the bug: it executes the parsed string through `/bin/sh` and asserts the resulting `PATH` has no parasitic quotes.

- **Before:** 14 passed, 15 failed, exit 1
- **After:** 29 passed, 0 failed, exit 0

Section 6 is a no-regression baseline of ordinary parsing — unquoted values, trailing comments, inline lists, `exclusive`, `verify_radius`, `depends_on`, an unquoted commit message containing a colon. It passed *before* the fix too, so it pins existing behaviour rather than describing the new one.

`repro/parse-real-cores.mjs` loads both shipped cores (`src/golden-cores/full-stack-app/core.yaml`, `src/golden-cores/landing-page/core.yaml`) through the real `loadCore` and dumps the parsed structure. The before-vs-after diff is **empty** — both parse identically. `node scripts/check.mjs` passes.

## One behaviour change worth flagging

A malformed scalar such as `verify: "echo hi" && test` previously returned the raw line including quotes; it now throws `unexpected content after a double-quoted scalar`. That is intentional — silent-wrong becomes loud — but it is a change. Neither shipped core contains such a pattern (the identical diff proves it), though an authored `.hedgehog/core.yaml` in the wild that relied on the old behaviour would now fail visibly at load time.

Not fixed, as out of scope: `parseInlineList` splits on `,` unconditionally, so a comma inside a quoted `scope` entry still breaks. Now documented in the header comment.

### DIFF
--- a/repro/parse-real-cores.mjs
+++ b/repro/parse-real-cores.mjs
@@ -1,0 +1,58 @@
+#!/usr/bin/env node
+// No-regression harness for the quoted-scalar escape fix.
+//
+// Loads every core.yaml that ships with (or lives in) this repo through
+// the real loader and prints a stable, fully-ordered dump of the parsed
+// result. Run it before and after a change to src/db/core.mjs and diff the
+// two outputs: identical output means ordinary core definitions parse
+// exactly as they did.
+//
+// Run: node repro/parse-real-cores.mjs
+
+import { readdirSync, statSync } from 'node:fs';
+import { join, relative } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { loadCore, parseCoreYaml } from '../src/db/core.mjs';
+import { readFile } from 'node:fs/promises';
+
+const root = fileURLToPath(new URL('..', import.meta.url));
+
+function findCoreYaml(dir, found = []) {
+  for (const entry of readdirSync(dir)) {
+    if (entry === '.git' || entry === 'node_modules') continue;
+    const full = join(dir, entry);
+    if (statSync(full).isDirectory()) findCoreYaml(full, found);
+    else if (entry === 'core.yaml') found.push(full);
+  }
+  return found;
+}
+
+const files = findCoreYaml(root).sort();
+if (files.length === 0) {
+  console.error('no core.yaml found — the harness would prove nothing');
+  process.exit(1);
+}
+
+let failed = 0;
+for (const file of files) {
+  const rel = relative(root, file);
+  console.log(`=== ${rel} ===`);
+  try {
+    // loadCore = parseCoreYaml + validateCore, the real entry point.
+    const core = await loadCore(file);
+    // Re-parse the raw text too, so a validation-only difference cannot
+    // hide a parse difference.
+    const raw = parseCoreYaml(await readFile(file, 'utf8'));
+    if (JSON.stringify(core) !== JSON.stringify(raw)) {
+      console.log('  !! loadCore and parseCoreYaml disagree');
+      failed++;
+    }
+    console.log(JSON.stringify(core, null, 2));
+  } catch (error) {
+    console.log(`  !! FAILED TO LOAD: ${error.message}`);
+    failed++;
+  }
+}
+
+console.log(`\n${files.length} core file(s), ${failed} failure(s)`);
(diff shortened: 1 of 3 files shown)


skyf0xx:
Most solid PR in the queue, all three defects reproduce against master's parser (`core.mjs:14-36`). `stripComment`'"'"'s truncation is the worst — silently cuts a verify command in half.

One call for you: malformed scalars now throw at load instead of returning raw silently — checked, neither shipped core hits this, but it's a behavior change worth a conscious yes.

No conflicts, clean rebase.

---

skyf0xx:

---

skyf0xx:
Landed on master at dd2d0a5, rebased onto current master (past 7450c84 and #29's layer-cardinality work). Clean rebase, no conflicts even in src/db/core.mjs despite #29 also modifying that file — the changes touched disjoint parts of parseScalar/stripComment vs validateCore/once.

Verified before landing: `node scripts/check.mjs` passes, all 29 assertions in this PR's own yaml-scalar-escapes.mjs pass (including running the parsed string through sh, the check that originally exposed the bug), parse-real-cores.mjs confirms both shipped Golden Cores — including #29's new `once` field — parse identically, and the full repro suite passes except the pre-existing, unrelated 05-shipped-cores-unchanged.mjs fixture-staleness failure.

Good find — a silent shell-quoting corruption that passes validation and only breaks at runtime is exactly the class of bug worth a real fix rather than a workaround.

---

PR #16, opened 2026-08-08
fix(verify): run git without a shell to stop command injection from file names and core.yaml

### LINKED ISSUE #15: [security] hedgehog verify executes shell code from file names and core.yaml fields
`hedgehog verify` builds shell command strings by interpolation and runs them through `execSync` (which uses `/bin/sh`). The values interpolated include **file names read from the working tree** and **fields from the project's `.hedgehog/core.yaml`**. Both execute.

Reported publicly because this repo has no `SECURITY.md`, no private advisory channel and no discussions enabled. The impact is local — the input comes from the project's own files and config, not from a network — so a public report seems proportionate. Happy to move it if you would rather have it private.

## Root cause

`src/db/verify.mjs`:

```js
function quotePathspec(path) {
  return JSON.stringify(path);
}
```

`JSON.stringify` escapes `"` and `\`. It does **not** escape `$` or backticks — and the result is placed inside double quotes in a shell command, where both are interpreted.

Affected call sites, all in `src/db/verify.mjs`:

| Line | Command | Interpolated value comes from |
|---|---|---|
| 78 | `git diff --name-only HEAD${scopeArgs}` and the adjacent `git ls-files --others` | `scope` globs in `core.yaml` |
| 250 | `git ls-tree -r --name-only HEAD -- ${quoted}` | file names in the working tree |
| 270 | `git add -- ${quoted}` | file names in the working tree |
| 271 | `git commit -m ${JSON.stringify(commitMessage)}` | the layer's `commit` field in `core.yaml` |

(`verify.mjs:236`, which runs the layer's own `verify_command`, is a shell command by design and is not part of this.)

## Confirmed by execution

**Vector 1 — a file name.** Creating a file called `src/$(touch OWNED_BY_FILENAME).ts` and running the same sequence `changedPaths` uses — read the names with `git ls-files --others --exclude-standard`, pass them through `quotePathspec`, interpolate into `git ls-tree` — produced the file `OWNED_BY_FILENAME`. The payload came from the name of a file sitting in the repository.

**Vector 2 — a commit message.**

```js
const commitMessage = "chore: pwn $(touch PWNED) `touch PWNED2`";
execSync(`git commit --allow-empty -m ${JSON.stringify(commitMessage)}`, { stdio: "pipe" });
```

Both `PWNED` and `PWNED2` were created, and the recorded commit subject was truncated to `chore: pwn` — so the injected text executes *and* disappears from the log.

## Why this matters more in Hedgehog than in a typical CLI

In Hedgehog's own model:

- **The files are written by AI agents.** `layer-eng` (and its per-core equivalents) creates files inside a task's scope, and `hedgehog verify` then shells out with their names.
- **`core.yaml` is authored by an AI agent** — `hedgehog-core-design` writes the layer definitions, including `commit` and `scope`.

So no external attacker is required for this to fire. A confused agent, or a prompt injection reaching one, is enough to turn a file name into a command executed by the gate — and the gate is the component the whole discipline trusts to be the thing that cannot be worked around.

It is also worth noting the failure is quiet. In vector 2 the injected text does not appear in the commit subject, so a log review after the fact shows nothing unusual.

## Suggested fix

Drop the shell at these call sites: use `execFileSync('git', [...args])`, which passes an argv array to the process directly, so no quoting is required and none can be got wrong. Pathspec semantics (including magic like `:(glob)…`) survive unchanged because git receives each argument verbatim.

A pull request doing exactly this, with a reproduction for each vector, follows.

## Elsewhere in the codebase

- `src/db/rebuild.mjs:57` — static command string, not affected.
- `src/golden-cores/full-stack-app/tools/phase-gate.cjs:19` — interpolates a `base` ref into `git log --format=%s ${range}`. Same shape, shipped into consuming projects; worth a look, not covered by the PR that follows.

---

Found while running a full 36-task authored-core build for an evaluation of the tool (see #8, #9, #10). It was not in the build's 40-entry friction log — it surfaced only when a subagent's report was checked against the code rather than taken at face value.

Intended labels: `bug`, `security` (filer has read-only access to this repo, so labels could not be applied at creation).

reinaldosaraiva:
Fixes #15.

`src/db/verify.mjs` built git command strings by interpolation and ran them through `execSync`, i.e. `/bin/sh`. The quoting helper was `JSON.stringify`, which escapes `"` and `\` but not `$` or backticks — and the result landed inside double quotes, where the shell interprets both. The interpolated values are file names read from the working tree and fields from the project's `core.yaml`.

## What changed

`execFileSync('git', [...])` at all four call sites, so git receives an argv array directly and no shell is involved:

- `git diff --name-only HEAD` + `git ls-files --others` (scope globs from `core.yaml`)
- `git ls-tree -r --name-only HEAD --` (file names)
- `git add --` (file names)
- `git commit -m` (the layer's `commit` message)

`scopeArgs` is now built by a `pathspecArgs()` helper returning `['--', ...pathspecs]`, preserving the degenerate empty-list case exactly. `git rev-parse HEAD` in the same function was converted too — a static string with zero behaviour change, but it removes the last needless shell from that path. `quotePathspec` is deleted.

`execSync` is retained for exactly one call site: `runVerifyCommand`, which runs the layer's own `verify` command. That one is a shell command by design and must stay one.

## It is also an integrity bug, not only an RCE

This was not in the original report and is worth stating plainly. Because the shell rewrites the pathspec, a payload-named file **silently evades `git add`**. Pre-fix, the verify commit contained only the innocent siblings; the file with the payload name stayed untracked while the task was marked `complete`.

So the gate certified a commit that did not contain the work it was gating. That is a failure of the property the whole discipline rests on, reachable without any intent to attack — any file whose name happens to contain shell metacharacters would do it.

The file-name vector also reaches **two** call sites, not one: the reproduction records four payload executions per malicious file name.

## Evidence

`repro/shell-injection.mjs` drives the real CLI (`db init` → `intent add` → `plan` → `claim` → `verify`) in a throwaway `mkdtemp` per scenario. Payloads only `touch`/`printf` a relative marker inside that temp dir.

**Before — 0/5 checks passed:**

```
Scenario A — scope glob        [FAIL] marker present
Scenario B — commit message    [FAIL] marker present
                               [FAIL] subject expected "feat(demo): layer $(touch ../pwn-commit)"
                                             actual   "feat(demo): layer"
Scenario C — file name         payload executions recorded: 4
                               files in the verify commit: src/evil.ts, src/tick.ts
                               [FAIL] marker present
                               [FAIL] payload-named files staged: expected 2, actual []
```

**After — 5/5:** zero executions, the subject recorded verbatim, and every in-scope file staged under its real name, payload-named ones included.

## No regression

`repro/normal-verify.mjs` — 23 assertions across 5 end-to-end scenarios: happy path (multi-glob scope, `:(glob)` pathspec semantics, created-vs-modified artifact rows read back from the DB, exact commit subject, exactly the in-scope files staged, clean tree afterwards); shell-awkward-but-legal file names (space, `;`, `&`, `()`, `'`, leading dash); a scope violation; a failing `verify_command` (`echo boom && exit 3`, which pins that `verify_command` is still interpreted by a shell); and an empty change set.

**23/23 on both the unfixed and the fixed code.** That is the load-bearing claim — the suite passes identically before and after, so it pins existing behaviour rather than describing the new one.

`node scripts/check.mjs` passes.

## Two things found along the way, deliberately not fixed here

- **Pre-existing and unrelated:** a file name containing `"` breaks verify entirely, before and after this change. `git ls-files --others` C-quotes such paths, so verify reads back the literal `"src/has\"quote.ts"` including the outer quotes, and the following `git add` fails with `pathspec ... did not match any files`. That is a path-decoding bug needing its own fix; it is documented in a comment in `repro/normal-verify.mjs`.
- **`src/golden-cores/full-stack-app/tools/phase-gate.cjs:19`** — `execSync(\`git log --format=%s ${range}\`)`, where `range` derives from `PHASE_GATE_BASE` / `GITHUB_BASE_REF` / `'origin/main'`. Same class, lower severity: git refnames legally permit `$`, backticks, `;`, `&`, `(` and `)`, so a branch name is a viable payload, though creating one in the base repo needs push access. It is also entirely unquoted — a space in `base` splits into two arguments — and lacks a `--` separator, so a crafted value could inject git options too. This file ships into every full-stack-app project's CI. The fix is the same shape: `execFileSync('git', ['log', '--format=%s', range])`. Left out of this PR to keep it to one concern; happy to add it here if you would rather.

### DIFF
--- a/repro/fixture.mjs
+++ b/repro/fixture.mjs
@@ -1,0 +1,150 @@
+// Shared fixture for the repro scripts: a throwaway git repo with a
+// one-layer authored core, an intent compiled into a task, and that task
+// claimed — the exact state a build agent hands to `hedgehog verify`.
+//
+// Everything lives inside a fresh `mkdtemp` directory. Nothing outside it
+// is ever written or removed.
+
+import { execFileSync } from 'node:child_process';
+import { mkdtempSync, mkdirSync, writeFileSync, readFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, resolve, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+export const CLI = resolve(HERE, '..', 'bin', 'cli.mjs');
+
+export function git(cwd, args) {
+  return execFileSync('git', args, { cwd, encoding: 'utf8', stdio: 'pipe' });
+}
+
+// Runs the real `hedgehog` CLI. A nonzero exit is a legitimate outcome
+// (a scope violation or a failing verify_command is exit 1), so it is
+// captured rather than thrown.
+export function hedgehog(cwd, args) {
+  try {
+    const stdout = execFileSync('node', [CLI, ...args], { cwd, encoding: 'utf8', stdio: 'pipe' });
+    return { status: 0, stdout, stderr: '' };
+  } catch (err) {
+    return {
+      status: err.status ?? 1,
+      stdout: err.stdout ?? '',
+      stderr: err.stderr ?? String(err.message),
+    };
+  }
+}
+
+// `hedgehog plan` spawns a detached graph-viewer server per project.
+// These fixtures don't want one, and a repro run that leaves a dozen
+// stray node processes behind is bad manners.
+function stopGraphServer(repo) {
+  const pidfile = join(repo, '.hedgehog', 'graph-server.json');
+  // The detached server writes its own pidfile, which can land a moment
+  // after `plan` returns — poll briefly rather than racing it.
+  const deadline = Date.now() + 3000;
+  while (Date.now() < deadline) {
+    try {
+      const { pid } = JSON.parse(readFileSync(pidfile, 'utf8'));
+      if (pid) {
+        process.kill(pid);
+        return;
+      }
+    } catch {
+      // Not written yet, half-written, or the server is already gone.
+    }
+    Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, 50);
+  }
+}
+
+// `scope` is spliced into core.yaml's inline-list syntax verbatim, so
+// callers pass e.g. `src/**` or `src/**, docs/**`. `seed` is a map of
+// repo-relative path -> contents committed before the task is claimed.
+export function makeProject({ scope, verify = 'true', commit, seed = {} }) {
+  const root = mkdtempSync(join(tmpdir(), 'hedgehog-verify-repro-'));
+  const repo = join(root, 'repo');
+  mkdirSync(join(repo, 'src'), { recursive: true });
+  mkdirSync(join(repo, '.hedgehog'), { recursive: true });
+
+  git(repo, ['init', '-q']);
+  git(repo, ['config', 'user.email', 'repro@example.invalid']);
+  git(repo, ['config', 'user.name', 'repro']);
+  git(repo, ['config', 'commit.gpgsign', 'false']);
+
+  // The build graph, the commit lock and the graph-viewer pidfile are
+  // engine state, gitignored in every real Hedgehog project. The pidfile
+  // especially: the detached server rewrites and finally removes it on
+  // its own schedule, so a tracked copy shows up as a spurious
+  // working-tree change in whichever scenario happens to race it.
+  writeFileSync(
+    join(repo, '.gitignore'),
+    '.hedgehog/hedgehog.db*\n.hedgehog/commit.lock\n.hedgehog/graph-server.json\n',
+  );
+  writeFileSync(
+    join(repo, '.hedgehog', 'core.yaml'),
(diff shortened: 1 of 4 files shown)


skyf0xx:
Superseded — fixed on master in 7450c84 (4.0.4), reimplemented from your diagnosis rather than merged. Policy here: contributor branches aren't checked out or run locally, so this was rebuilt and verified independently rather than merged as-is.

`phase-gate.cjs:19` had the same bug and is fixed in the same commit — you flagged it as out of scope, it wasn't.

Closing as superseded. Thanks — the diagnosis did the real work.

---

skyf0xx:
Correction: I said `repro/` was excluded because it'"'"'d ship to the package. Wrong — `files[]` in package.json already excludes it, verified with `npm pack --dry-run`. Repo-tidiness call, not a supply-chain one. The decision to reimplement stands for the other reason given.

---

PR #17, opened 2026-08-08
feat(plan): reconcile compiled tasks with core.yaml, and surface drift instead of hiding it

reinaldosaraiva:
Editing `.hedgehog/core.yaml` after `hedgehog plan` has no effect on already-compiled tasks, and nothing warns. This adds a reconciliation path and, more importantly, makes the divergence visible.

## Confirmed in the code

`plan.mjs` copies the layer straight onto each task row:

```js
scope_globs: JSON.stringify(layer.scope.map((g) => fillModule(g, module))),
verify_command: fillModule(layer.verify, module),
commit_message: fillModule(layer.commit, module),
```

`schema.mjs` makes those real columns, and `plan` only ever INSERTs — `insertTask` is the only statement touching them, with no UPDATE path anywhere. `planTasks` skips an intent whose first task exists, and it only loads `PENDING_INTENT_STATUSES = ['proposed', 'planned']` while compiling flips the intent to `active`. So re-running `plan` after a `core.yaml` edit prints `0 intent(s) compiled` and the task keeps the stale command — reproduced literally.

On the build where this surfaced, a PATH fix to one layer's `verify` left **0 of 36 tasks** carrying the corrected version, and the only remedy was a direct `UPDATE` in SQLite.

## What this adds

- **`src/db/drift.mjs`** (new) — `detectDrift` / `recompileTasks` / `formatDrift` / `formatRecompile`. It separates three cases deliberately: field drift (rewritable), a layer dropped from `core.yaml`, and a changed `depends_on` edge. The last is graph shape, so it is reported and never silently rewritten.
- **`hedgehog plan --recompile`** — rewrites `objective`, `scope_globs`, `verify_command`, `commit_message`, `exclusive` and `verify_radius` on `proposed`/`planned`/`ready` tasks. It refuses `building`/`verifying`/`complete`/`blocked`, printing each skipped task with its reason and both values. Flags: `--dry-run`, `--include-blocked`, `--strict` (non-zero exit while drift remains). Status is re-checked **inside** the write transaction, so a task claimed mid-run is never rewritten under a live lease.
- **`hedgehog status`** — a new DRIFT section, field by field, pointing at `plan --recompile`. It degrades gracefully when there is no `core.yaml` or it will not parse.
- **Plain `hedgehog plan`** now warns about drift instead of reporting "0 compiled" and looking like a successful no-op.
- **`hedgehog db rebuild`** reports post-rebuild drift, and states that hand-edited task rows have no committed source and are not replayed.
- **`refactor(plan)`** extracts `layerTaskFields()` + `LAYER_DERIVED_FIELDS`, so the drift check compares exactly the columns the compiler writes. No behaviour change.

## A documentation correction

`src/templates/CLAUDE.core.authored.md` told the operator to treat `core.yaml` as authoritative when it and the compiled task disagree — the exact opposite of the real behaviour, which is what made this bug expensive to diagnose. Corrected to describe what actually happens, and to point at `--recompile` as the way to make the two agree.

## Two corrections to the original diagnosis

Worth recording, since both make the picture sharper:

- **`rebuild` is not only a fresh-clone concern.** It calls `planTasks` with the *current* `core.yaml`, so a rebuild does pick up `core.yaml` edits for whatever it compiles fresh — which is precisely why hand-patched task rows vanish. Verified by deleting the DB and rebuilding: the new commit message from the edited `core.yaml` appeared.
- **On a module-axis core, widening a layer is not merely the worse workaround — it is the only representable one.** `validateCore` actively forces uniformity:

  ```js
  if (!layer.scope.join('').includes('{module}')) {
    throw new Error(`layer "${layer.id}" has no {module} in scope, but core "${core.id}" is module-axis ...`);
  }
  ```

## Evidence

`node repro/run-all.mjs` — 3 scripts, 11 scenarios, driving the real CLI in throwaway temp dirs.

- **Before** (master, via a throwaway worktree): 0/3 scripts, 9 of 11 scenarios fail — e.g. `expected: "PATH=/usr/local/bin:$PATH node --test infra/billing"` / `actual: "node --test infra/billing"`, and `expected output to contain: "DRIFT"` against a status output with none.
- **After:** 3/3 scripts, 11/11 scenarios pass.

`node scripts/check.mjs` passes. `src/db/verify.mjs` is untouched.

## On the per-module scope exception — a recommendation, not an implementation

The related half of this problem is that a per-module scope exception is not expressible at all, so real builds patch individual task rows via SQL and the graph diverges from its own committed sources.

A committed record of task-level corrections would fit the existing shape almost exactly, because `.hedgehog/intents/*.json` already establishes the pattern: committed, plain text, replayed by `rebuild` through `planTasks`. Something like `.hedgehog/overrides/*.json` with entries of the form `{ task: "BILLING-FOUNDATION", scope_add: ["infra/shared/registry.ts"], reason: "..." }`, applied by `planTasks` immediately after insert — a fresh compile applies it, `rebuild` reapplies it, a PR can review it, and `hedgehog why` can reach it.

Three constraints seem worth holding firm on: **additive only** (`scope_add`, never `scope_replace`, so an override can never shrink a gate); **per task, never per layer** (a per-layer override is a layer edit wearing a disguise); and **no `verify` overrides** at all, since weakening a layer's verify is the one thing the discipline forbids outright — that belongs in `core.yaml` plus `--recompile`, which is now a supported path.

This is deliberately **not implemented here**, because it creates a second authority over `scope_globs` — the same class of problem as the bug this PR fixes. Adopting it means `detectDrift` must compare against `core.yaml` *composed with* the overrides, or every overridden task reads as permanently drifted and the new DRIFT section becomes noise. That coupling makes it your design call rather than a patch.

## Honest gaps

- `--include-blocked` is implemented but has **no repro assertion** — the reproduction covers `building`/`verifying`/`complete` only.
- Dependency-edge drift detection and the rebuild drift warning are implemented and manually smoke-tested, **not** covered by an automated reproduction.
- The rebuild warning does not compare against "what it previously held" in the fresh-clone case, because there is nothing to compare against — the DB is gitignored and `planTasks` is additive. It reports post-rebuild disagreement with `core.yaml` instead, which is useful on a populated DB.
- The reproductions drive the real CLI for everything except task lifecycle transitions, which are written directly to the status/lease columns — a real `verify` needs a toolchain the throwaway project does not have. The fields under test are only ever written by the CLI.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -26,6 +26,7 @@ import { verifyTask } from '../src/db/verify.mjs';
 import { claimTasks, releaseTask, renewLease } from '../src/db/claim.mjs';
 import { readyTasks, formatReady } from '../src/db/ready.mjs';
 import { graphStatus, formatStatus } from '../src/db/status.mjs';
+import { detectDrift, recompileTasks, formatRecompile } from '../src/db/drift.mjs';
 import { whyPath, formatWhy } from '../src/db/why.mjs';
 import { addFriction, listFriction } from '../src/db/friction.mjs';
 import { rebuildDb } from '../src/db/rebuild.mjs';
@@ -221,6 +222,22 @@ async function ensureDb() {
   console.log(
     `${dim('DB missing — rebuilt from')} ${bold(INTENTS_DIR)}${dim(':')} ${dim(`${result.intentsReplayed} intent(s) replayed, ${result.tasksMarkedComplete} task(s) marked complete`)}\n`,
   );
+  warnRebuildDrift(result, corePath);
+}
+
+// A rebuild re-derives every task from the current core.yaml and the
+// committed intents. Nothing else is committed — a task row someone
+// patched by hand has no source to replay from — so say plainly what a
+// rebuild can and cannot carry, and surface any task that ends up
+// disagreeing with core.yaml.
+function warnRebuildDrift({ drift }, corePath) {
+  if (!drift || drift.length === 0) return;
+  console.log(
+    `${yellow(bold('Core drift after rebuild.'))} ${drift.length} task(s) do not match ${bold(corePath)}.\n` +
+      `A rebuild replays ${bold(INTENTS_DIR)} and re-derives tasks from core.yaml; task rows\n` +
+      'edited by hand have no committed source and are not replayed. Run\n' +
+      `${bold('hedgehog status')} for the divergence, ${bold('hedgehog plan --recompile')} to reconcile.\n`,
+  );
 }
 
 // Writes one planned file to disk — a straight copy, or for a `merge`
@@ -307,6 +324,8 @@ ${bold('Usage')}
   npx @skyf0xx/hedgehog db rebuild                re-derive the build graph from committed intents + git history
   npx @skyf0xx/hedgehog plan                      compile pending intents into tasks + dependencies,
                                                    then open the build graph if anything compiled
+  npx @skyf0xx/hedgehog plan --recompile          rewrite core.yaml-derived fields on not-started tasks
+                                                   [--dry-run] [--include-blocked] [--strict]
   npx @skyf0xx/hedgehog intent add [flags]        add an intent (rules/requirements/dependencies)
   npx @skyf0xx/hedgehog intent add --file <path>  add an intent from a JSON file
   npx @skyf0xx/hedgehog next                      print the task packet for one ready task
@@ -314,7 +333,8 @@ ${bold('Usage')}
   npx @skyf0xx/hedgehog release <task-id> --owner <owner>   hand a claimed task back to ready
   npx @skyf0xx/hedgehog renew <task-id> --owner <owner> [--minutes <n>]   extend a held lease
   npx @skyf0xx/hedgehog verify <task-id> --owner <owner>   run scope + verify checks, commit on pass
-  npx @skyf0xx/hedgehog status                    graph overview: counts by status, ready list, in flight
+  npx @skyf0xx/hedgehog status                    graph overview: counts by status, ready list, in flight,
+                                                   and any drift from core.yaml
   npx @skyf0xx/hedgehog ready                     preview which ready tasks are claimable now vs held back
   npx @skyf0xx/hedgehog quiesce                   report whether anything is still in flight
   npx @skyf0xx/hedgehog graph                     start (or reuse) the live graph server and open it
@@ -532,6 +552,7 @@ async function dbRebuildCommand() {
   console.log(
     `${green('rebuilt')}  ${dim(`${result.intentsReplayed} intent(s) replayed, ${result.tasksMarkedComplete} task(s) marked complete`)}\n`,
   );
+  warnRebuildDrift(result, corePath);
 }
 
 async function dbCommand(args) {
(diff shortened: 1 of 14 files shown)


skyf0xx:
Confirmed, no gate weakening — `plan --recompile` only touches `proposed`/`planned`/`ready`, rewrites toward `core.yaml`, never toward looser.

**Sequencing issue with #29**: your `detectDrift` calls `layerTaskFields(layer, task.module)`; #29'"'"'s once-layers use synthetic module `_core`, which won'"'"'t match a scope with no `{module}`. Land in the wrong order and once-tasks report permanent drift. Whoever lands second needs a hand-merge, not a rebase.

Prerequisite for #26, not a fix for it — noted there.

---

skyf0xx:
This branch is based on `2bcc5c4`, before `7450c84` landed on master (a shell-injection fix in `src/db/verify.mjs`). This PR touches `src/db/plan.mjs`, not `verify.mjs` directly, so there's no injection concern here — but please rebase onto current `master` so the merge base is current before this goes in.

---

skyf0xx:
Landed on master at f799f3d, rebased onto current master. You flagged this conflict explicitly in #29's own PR body — "near-certain, and substantive rather than textual" — and that call was right, but the mechanical rebase itself was actually clean: layerTaskFields() coexists with compileOnceTasks unmodified, and the substantive part turned out to be a gap the two features left between them that neither PR's diff could have surfaced on its own.

The mechanical conflicts (src/db/plan.mjs's field-extraction hunk, bin/cli.mjs's imports, the repro/lib.mjs+run-all.mjs filename collision — renamed to drift-lib.mjs, same pattern as every other PR in this queue) all resolved by keeping both sides' logic side by side.

The real finding, from actually running a once-layer core through detectDrift rather than trusting the rebase: taskDrift() called layerTaskFields(layer, task.module) unconditionally, including for once-layer tasks. A once-task's module is CORE_MODULE ('_core'), so layerTaskFields would substitute '_core' into templates that compileOnceTasks never substitutes at all — every once-layer task read as permanently, unfixably drifted the moment #29 and this PR coexisted. A second, subtler bug rode along: the dependency-edge comparison assumed every parent shares the task's own intent, which breaks for the two cross-cardinality edge shapes #29 documented (per-intent → once, and once → per-intent, where a tail once-task's expected parent set is one task per intent that compiled the parent layer — genuinely not derivable from core.yaml alone, only from what's actually been planned).

Fixed by extracting a once-layer counterpart to layerTaskFields (onceLayerTaskFields, in plan.mjs, used by both compileOnceTasks and the drift check — same 'single definition' principle you established for the per-intent case), and branching taskDrift's field and dependency comparisons on task.module === CORE_MODULE. Verified directly: a two-intent, once-before/once-after core (cluster → schema → deploy) now reports zero drift immediately after a fresh plan, in both directions, across all four depends_on cardinality combinations.

Verified before landing: `node scripts/check.mjs` passes, the full 20-script numbered suite passes except the pre-existing, unrelated 05-shipped-cores-unchanged.mjs fixture-staleness failure, all of this PR's own reproductions pass, and #29's own reopen-for-new-scope reproduction (08) still passes — confirming the fix doesn't interfere with re-entry.

Nicely scoped PR otherwise — good discipline keeping the per-task override design out of this one rather than letting scope creep in.

---

PR #18, opened 2026-08-08
feat(cli): add hedgehog boundary, so the discipline can answer when to clear context

reinaldosaraiva:
The discipline's central claim is that the conversation is disposable — clear context at a natural boundary and recover from the graph. The template says exactly that, and tells the operator to run `hedgehog quiesce` first. But `quiesce` answers only whether anything is in flight, which is a **necessary condition, not a sufficient one**, so nothing in the tool can actually answer the question the discipline asks.

## What was missing

A boundary needs three things the tool already knows separately:

1. nothing is in flight,
2. the working tree is clean,
3. the last closed task completed an intent.

`hedgehog verify` already prints `intent complete` when (3) happens — `completeIntentIfDone` in `verify.mjs` — and then discards it. Nothing stores or re-derives it. So the boundary decision lives in the operator's head, which is precisely the kind of state the discipline says it does not want anyone holding.

The second half is positioning: when a fresh session starts, nothing tells it where the build is. On the project where this surfaced, a hand-maintained handoff file was the only thing doing that job.

## Interface

```
hedgehog boundary [--quiet] [--handoff]
```

- **Exit 0** — boundary reached. **1** — not a boundary. **2** — unanswerable (no build graph, no git repo, or complete tasks exist but the last-closed one cannot be identified).
- **stdout** is the capturable payload: the `NEXT`/`WHY` block on success, or `--handoff`'s full block (BUILD + HEAD, boundary verdict, NEXT/WHY, ALLOWED SCOPE, VERIFICATION, IN FLIGHT, BLOCKED, BLOCKED DOWNSTREAM).
- **stderr** is the human verdict: all three conditions with ✓/✗/? and a detail line each.
- `--quiet` gives the exit code only, for a shell hook. `--handoff` prints its block regardless of verdict — a fresh session needs positioning either way — while keeping the exit code honest.

`ensureDb` gained a `log` sink so its rebuild notice goes to stderr for this one command; every other command is unchanged. `quiesce` is untouched.

At a genuine boundary:

```
stderr: Boundary reached.
          ✓ nothing in flight — no task building or verifying
          ✓ working tree clean — no uncommitted changes
          ✓ last closed task completed an intent — last closed task ALPHA-VIEW completed intent "alpha"
stdout: NEXT
          BETA-MODEL   model   model for beta
        WHY
          intent "beta" — beta goal ... 1 task(s) blocked downstream of it
```

Mid-intent, `quiesce` exits 0 while `boundary` exits 1 with:

```
✗ last closed task completed an intent — last closed task ALPHA-MODEL left intent "alpha" open — 1 task(s) remain: ALPHA-VIEW
```

That contrast is the whole point of the command.

## One correction to the premise

There is no `completed_at` column and no verify timestamp, so "the last closed task" is not directly stored. It is derived from the newest `passed` row in `verifications`, with a fallback that matches git commit subjects against `tasks.commit_message`.

The fallback is necessary because `hedgehog db rebuild` reconstructs `tasks.status` from commit subjects and writes **no** verification rows and **no** `intents.status`. For the same reason, intent closure is tested with `completeIntentIfDone`'s own predicate rather than by reading `intents.status`, which is simply wrong on a rebuilt graph. Both paths are covered.

## Evidence

`node repro/run-all.mjs` — four scripts, each building a real temp project (git repo, authored `core.yaml`, two intents, real claim/verify/commit cycles) and driving `bin/cli.mjs`.

- **Before** (`bin/cli.mjs` stashed to pristine master): all four fail — e.g. `expected to contain: "✗ last closed task completed an intent"` / `actual: "Unknown command: boundary"`.
- **After:** all four pass.

Worth noting how the reproductions are written: the exit-code assertions alone would pass *vacuously* against master, because an unknown command also exits 1. So each reproduction additionally asserts that the failing condition is **named**. A test that passes for the wrong reason is worse than no test.

`node scripts/check.mjs` passes. `src/db/verify.mjs`, `plan.mjs` and `next.mjs` are untouched — `boundary.mjs` imports `nextTask`/`stalledTasks`/`graphStatus` read-only.

## Documentation

`src/templates/CLAUDE.md`, `src/skills/hedgehog-loop/SKILL.md` and `hedgehog-authored-loop/SKILL.md` told the operator to use `quiesce` for this. They now point at `boundary` and explain the difference between "nothing in flight" and "this is a boundary".

## Honest caveats

- Condition 3 is fully reliable on a live graph. On a **rebuilt** graph it depends on commit subjects being unique per task — and a core whose `commit:` template omits `{module}` can make two complete tasks share a subject. In that case the command reports the ambiguity and exits 2 rather than guessing.
- A task whose verify touched no in-scope files makes no commit at all, so on a rebuilt graph it is invisible to the fallback. The fallback scans up to 200 subjects and says plainly when none match. (That underlying gap is the subject of a separate PR.)
- Working-tree cleanliness excludes the engine's own derived files (`.hedgehog/hedgehog.db*`, `commit.lock`, `graph-server.json`), mirroring `verify.mjs`'s `isEngineStatePath`.
- `README`/`ARCHITECTURE` were left alone — neither enumerates these graph commands.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -26,6 +26,7 @@ import { verifyTask } from '../src/db/verify.mjs';
 import { claimTasks, releaseTask, renewLease } from '../src/db/claim.mjs';
 import { readyTasks, formatReady } from '../src/db/ready.mjs';
 import { graphStatus, formatStatus } from '../src/db/status.mjs';
+import { boundaryState, formatBoundary, formatPosition, formatHandoff } from '../src/db/boundary.mjs';
 import { whyPath, formatWhy } from '../src/db/why.mjs';
 import { addFriction, listFriction } from '../src/db/friction.mjs';
 import { rebuildDb } from '../src/db/rebuild.mjs';
@@ -196,7 +197,11 @@ const exists = (p) =>
 // there's nothing to rebuild from either (a genuinely fresh project, no
 // intents yet), this no-ops and leaves the DB missing — the caller's own
 // existing "No build graph found" guard still fires for that case.
-async function ensureDb() {
+// `log` exists for the one caller whose stdout is a machine-consumable
+// payload (`hedgehog boundary`): the rebuild notice is commentary, and
+// commentary on stdout would corrupt a block meant to be captured or
+// piped. Every other command leaves it at the default.
+async function ensureDb({ log = console.log } = {}) {
   if (await exists(DB_PATH)) return;
 
   let intentFiles = [];
@@ -218,7 +223,7 @@ async function ensureDb() {
   } finally {
     db.close();
   }
-  console.log(
+  log(
     `${dim('DB missing — rebuilt from')} ${bold(INTENTS_DIR)}${dim(':')} ${dim(`${result.intentsReplayed} intent(s) replayed, ${result.tasksMarkedComplete} task(s) marked complete`)}\n`,
   );
 }
@@ -317,6 +322,10 @@ ${bold('Usage')}
   npx @skyf0xx/hedgehog status                    graph overview: counts by status, ready list, in flight
   npx @skyf0xx/hedgehog ready                     preview which ready tasks are claimable now vs held back
   npx @skyf0xx/hedgehog quiesce                   report whether anything is still in flight
+  npx @skyf0xx/hedgehog boundary                  is this a moment to clear context? exits 0 only if it is,
+                                                   and prints what a fresh session picks up next
+  npx @skyf0xx/hedgehog boundary --handoff        print the block a fresh session starts from
+  npx @skyf0xx/hedgehog boundary --quiet          exit code only, for a shell hook
   npx @skyf0xx/hedgehog graph                     start (or reuse) the live graph server and open it
   npx @skyf0xx/hedgehog graph --no-open           start (or reuse) the server; print the URL instead
   npx @skyf0xx/hedgehog why <path>                provenance chain for a file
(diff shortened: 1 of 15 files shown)


skyf0xx:
Real gap (`quiesce` only answers about in-flight work), clean implementation, minimal conflicts.

One request: `boundary.mjs`'"'"'s `git log`/`git status` calls use `execSync` with string commands. Not injectable as written, but 7450c84 moved the whole codebase off shell-based git — please use `execFileSync` here too for consistency.

---

skyf0xx:

---

skyf0xx:
Landed on master at d9796a3, rebased onto current master. One conflict: both this PR and #29 independently wrote repro/run-all.mjs. Kept #29's glob-based version (auto-discovers numbered repro scripts, consistent with what #11/#12/#13/#14 already use) rather than this PR's hardcoded boundary-*.mjs list — your four boundary-*.mjs scripts aren't digit-prefixed so they don't match the glob, but they run standalone fine (verified individually below).

Verified before landing: `node scripts/check.mjs` passes, all 4 of this PR's own reproductions pass (boundary-at-boundary, boundary-work-in-flight, boundary-dirty-tree, boundary-intent-incomplete), and the full numbered repro suite passes except the pre-existing, unrelated 05-shipped-cores-unchanged.mjs fixture-staleness failure. This command doesn't touch verify.mjs, plan.mjs, or any of the once-layer logic, so no interaction with #29 to re-verify beyond the file collision above.

---

PR #19, opened 2026-08-08
feat(packet): carry the intent goal to every builder, restate it when the intent closes, and let tasks declare debt

reinaldosaraiva:
A layer can pass its own verify without building what the intent asked for. On the build where this surfaced, an intent whose `goal` began *"allow creating, editing and moving Cards between Lists"* compiled into a `domain-service` task. The agent built `create`, `moveToList` and `reorderWithinList` — and no `update` of the Card's scalar fields. It passed with **117 green tests** and was committed. The gap was caught two layers later, by a human, not by any gate.

## ⚠️ Overlap with #12 — please read first

One commit here (`feat(packet)`) makes `hedgehog claim` emit the full packet. **#12 does the same thing**, as part of its recovery-commands work. The two were written independently and will conflict textually.

If #12 lands first, `feat(packet)`'s claim change becomes redundant and I will rebase this branch to drop it — the rest of this PR (the `GOAL`/`OUTCOME` labels, the intent check, the debt table) is independent and still applies. Happy to do that pre-emptively if you would rather review them in a fixed order.

## A correction to my own diagnosis

I opened this expecting the packet not to carry the intent's goal. **That was wrong**, and the reproduction says so explicitly: `formatNext` in `src/db/next.mjs` already printed it —

```js
lines.push('INTENT');
lines.push(`  ${intent.goal}`);
lines.push(`  ${intent.outcome}`);
```

The real hole is one surface over. `hedgehog claim` — which `templates/CLAUDE.md` and every loop skill tell agents to dispatch from — printed **no packet at all**. Against unmodified master, the entire output is:

```
Claimed. Task CARD-DOMAIN-MODEL leased to sess1.
  expires 2026-08-08 21:38:07
```

No goal, no rules, no scope, no verification. Meanwhile `src/skills/hedgehog-authored-loop/SKILL.md:61` and `src/templates/CLAUDE.md:72` both assert that claim "returns up to N task packets (STATUS/INTENT/RELEVANT RULES/WHY NOW/… each)". **The documentation described a packet the CLI never emitted**, and that is the whole explanation for the observed bug: the agent building that layer genuinely never saw the goal, because the command it was dispatched from never printed one.

## What this adds

- **`feat(packet)`** — labels the two INTENT lines `GOAL`/`OUTCOME`, and makes `claim` emit the full packet per claimed task (via an exported `assemblePacket`). *(This is the commit that overlaps #12.)*
- **`feat(verify)`** — `completeIntentIfDone` now returns the intent row rather than a bare boolean, and when the **last** layer of an intent closes, the CLI prints an `INTENT CHECK` block restating the goal and outcome. The point is not automation — it cannot be automated — but that the comparison between *what was asked* and *what was built* exists **at least once**, at the moment the intent is claimed to be done. The change to `verify.mjs` is 23 added / 6 removed, confined to that one function and the return object.
- **`feat(debt)`** — a `debt` table, `hedgehog debt add <task-id> "<note>"` / `debt list [<task-id>]`, and an `INHERITED DEBT` packet section fed by the declaring task's **transitive** downstream. This addresses the related failure where a layer discovered a real limitation, wrote it into a code comment under a "KNOWN LIMITATION" heading, and the task that inherited the problem was never told. A comment in a file is not a mechanism.

Two deliberate choices in the debt work:

- **No committed markdown log**, unlike `friction`. A file written under `.hedgehog/` mid-task is outside every task's scope globs and would trip verify's scope gate on the very task declaring the debt. The cost is that debt does not survive `hedgehog db rebuild`.
- `dbInit` now applies the schema unconditionally (everything is `IF NOT EXISTS`) so pre-existing graphs pick up the table, and the packet's debt read tolerates the table being absent — verified by dropping it and running `next`.

## Evidence

Repros under `repro/` — a `lib.mjs` harness plus 3 scripts, driving the real CLI against a throwaway git repo in a temp dir.

- **Against exported master:** all 3 exit 1 — `FAIL claim: emits the full packet, not just a lease line`; `FAIL closing the last layer frames the restatement as a check`; `FAIL hedgehog debt add is not a command`. The harness *also* confirms the two `next: packet carries the intent goal/outcome` assertions already passed on master — the correction above, made visible in the reproduction's own output rather than only claimed in prose.
- **Against this branch:** 3/3 pass. `node scripts/check.mjs` passes.

`hedgehog next` output diff on an unchanged fixture — this is the complete change:

```diff
 INTENT
-  allow creating, editing and moving Cards between Lists
-  a Card can be created, its scalar fields updated, and moved between Lists
+  GOAL     allow creating, editing and moving Cards between Lists
+  OUTCOME  a Card can be created, its scalar fields updated, and moved between Lists
 
 RELEVANT RULES
   - moving a Card between Lists restarts the time-in-list clock
 
+INHERITED DEBT
+  (none declared)
+
 WHY NOW
```

One section added; two existing lines gained labels with their text byte-identical. Nothing removed, renamed, reordered or reworded.

Payload docs updated to match: `templates/CLAUDE.md`, `hosts/routing.mjs`, `hedgehog-loop`, `hedgehog-authored-loop`, `hedgehog-landing-loop`, `layer-eng`, `backend-eng`, `front-end-eng`.

## Honest caveats

- The `INTENT CHECK` is text a human has to read. If nobody reads verify's stdout it changes nothing — it is not, and cannot be, a gate.
- Debt is graph-only and lost on `hedgehog db rebuild`. A durable version needs a structured committed file plus a replay path in `rebuild.mjs`, and would need `isEngineStatePath` widened so writing it mid-task does not trip the scope gate. That is more than "a new table plus packet rendering", so I stopped short.
- `formatNext` still hardcodes `STATUS   READY` even when emitted from `claim`, where the task is `building`. Deriving it from `task.status` would change existing `next` output, since `plan.mjs` inserts tasks as `planned`. Left alone deliberately.
- **Incidental finding, not fixed:** `.hedgehog/intents/*.json` (written by `intent add`) and `.hedgehog/friction/log.md` (written by `friction add`) are outside every task's scope globs, so leaving either uncommitted trips verify's scope gate on an unrelated task. Hit while building the repro harness; pre-existing and outside this PR.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -821,6 +821,27 @@ async function verifyCommand(args) {
   if (result.intentComplete) {
     console.log(`  ${green('intent complete')}  ${dim('every task for this intent is done')}`);
   }
+
+  // The one comparison in the circuit. Each layer's verify_command runs
+  // the tests that layer itself wrote, so it measures internal
+  // consistency and never coverage of what was asked — a layer that
+  // builds half an intent and tests that half exhaustively is green.
+  // Closing the last layer is the moment the intent is claimed done, so
+  // that is where what was requested gets printed back to be read against
+  // what was built. It is not a machine-checkable gate; it cannot be. Its
+  // value is that the comparison happens at all, once.
+  if (result.completedIntent) {
+    const { id, goal, outcome } = result.completedIntent;
+    console.log('');
+    console.log(`${bold('INTENT CHECK')}  ${bold(id)}  ${dim('— this closed the last layer of this intent.')}`);
+    console.log(`  ${dim('GOAL')}     ${goal}`);
+    console.log(`  ${dim('OUTCOME')}  ${outcome}`);
+    console.log('');
+    console.log(dim('  Confirm the work built across this intent\'s layers covers the above.'));
+    console.log(dim('  Anything asked for and not built is a Correction Protocol case now,'));
+    console.log(dim('  not a later discovery. Nothing else in the build checks this.'));
+    console.log('');
+  }
 }
 
 // `hedgehog claim --owner <owner> [--count <n>]` — atomically claims up to
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -21,7 +21,7 @@ import { dbInit, DB_PATH, openDb } from '../src/db/init.mjs';
 import { loadCore } from '../src/db/core.mjs';
 import { planTasks } from '../src/db/plan.mjs';
 import { addIntent, INTENTS_DIR } from '../src/db/intent.mjs';
-import { nextTask, formatNext, stalledTasks } from '../src/db/next.mjs';
+import { nextTask, formatNext, stalledTasks, assemblePacket } from '../src/db/next.mjs';
 import { verifyTask } from '../src/db/verify.mjs';
 import { claimTasks, releaseTask, renewLease } from '../src/db/claim.mjs';
 import { readyTasks, formatReady } from '../src/db/ready.mjs';
@@ -825,8 +825,14 @@ async function verifyCommand(args) {
 
 // `hedgehog claim --owner <owner> [--count <n>]` — atomically claims up to
 // `count` mutually non-conflicting ready tasks (claimTasks's fan-out, item
-// 13) and prints each one's packet-level summary, plus which owner now
-// holds them.
+// 13) and prints each one's FULL packet, plus which owner now holds them.
+//
+// The full packet, not a summary: `claim` is the surface the loop skills
+// dispatch from ("returns up to N task packets ... each"), so an agent
+// handed only a task id and a lease expiry never sees the intent's goal,
+// its rules, its scope, or its verification — it sees a layer name and
+// builds whatever that name suggests. `hedgehog next` prints the same
+// thing for the read-only case; the two now agree.
 async function claimCommand(args) {
   await ensureDb();
 
@@ -848,8 +854,13 @@ async function claimCommand(args) {
 
   const db = openDb();
   let claimed;
+  let packets = [];
   try {
     claimed = claimTasks(db, { owner, count });
+    // Assembled on the same open handle, right after the claim: the
+    // packet is what the caller dispatches, so it has to come back from
+    // the same call that handed out the lease.
+    packets = claimed.map((task) => assemblePacket(db, task));
   } finally {
     db.close();
   }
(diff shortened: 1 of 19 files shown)


skyf0xx:
Clean, no gate weakening. No conflict with 7450c84 despite touching verify.mjs — different regions, confirmed by reading both. Re-diff after rebasing anyway since you'"'"'re the only PR touching that file.

Overlaps #27 (`formatNext` tail) and #12 (packet-claim, you'"'"'ve offered to drop your duplicate).

---

skyf0xx:
This branch is based on `2bcc5c4`, before `7450c84` fixed a shell-injection path in `src/db/verify.mjs` (string-interpolated `execSync` → argv-based `execFileSync`). This PR also touches `verify.mjs`, so a mechanical rebase isn't enough — please rebase onto current `master` and re-check that your changes still use the argv-based `git()` helper rather than reintroducing any shell-string interpolation. GitHub will show this as a clean merge either way, so the safety of the merge depends on this being done deliberately, not just on conflict resolution.

---

skyf0xx:
Landed on master at 110dbe4, rebased onto current master. This was the most involved rebase in the queue — thank you for flagging the #12 overlap explicitly in the PR description, it made resolving it correctly much faster.

Since #12 landed first (as you anticipated), the redundant half of feat(packet) — making claim print packets via a new assemblePacket export — was dropped in favor of #12's existing printPackets/taskPacket mechanism, which already covers it (including for targeted claim, which this PR predates). Kept the genuinely new part: labelling the two INTENT lines GOAL/OUTCOME.

feat(verify) (the INTENT CHECK on intent close) and feat(debt) (the debt table, CLI commands, and INHERITED DEBT packet section) are both independent of #12 and landed as designed, merged alongside #12's incompleteDeps field in the same assemblePacket/formatPacket functions — both fields now coexist in the packet.

Two mechanical conflicts, not logic conflicts: repro/lib.mjs and repro/run-all.mjs collided with #12's identically-named files. Renamed this PR's harness to repro/packet-lib.mjs and repointed its three new scripts at it; kept #12's run-all.mjs (glob-based, already the convention every other landed PR uses).

One real bug from the rebase, caught before landing: the debt commit's 3-way auto-merge resurrected a dead assemblePacket(db, task) call in claimCommand — a leftover from the exact redundant code path described above, reintroduced because that commit's diff was based on the pre-fix file state. Running repro/run-all.mjs surfaced it immediately (4 of #12's own reproductions failed with "assemblePacket is not defined"); removed the dead code and folded the fix back into the debt commit so history stays clean.

Verified before landing: `node scripts/check.mjs` passes, all of this PR's own reproductions pass (packet-carries-intent-goal, intent-close-restates-goal, debt-reaches-inheriting-packet), and re-ran #11's, #12's, and #18's own reproductions too since this touched bin/cli.mjs and next.mjs — all still pass. Only the pre-existing, unrelated 05-shipped-cores-unchanged.mjs fixture-staleness failure remains.

---

Issue #1, opened 2026-07-31
landing-page core: delegated phase subagents may lack Bash tool access needed for "one phase = one commit"


skyf0xx:
## What happened

During a landing-page core build, two consecutive Chain Method phases
(systems, run by `landing-systems`, and sequence, run by
`landing-sequencer`) were delegated to subagent instances. Both
subagent sessions produced correct file-write output for their phase's
artifact, but neither had a Bash tool available in its session — so
neither could run `git add`/`git commit` for its own phase, as
`hedgehog-landing-loop` requires ("one phase = one commit," done as
part of the phase).

In both cases, the orchestrating session had to review the subagent's
output itself and perform the git add/commit mechanically, outside the
delegated context, after the fact.

## Why this matters

The discipline's model is that a phase isn't done until it's committed
by the agent that ran it — commit authorship and phase completion are
meant to be the same act. When a subagent can't commit its own work,
that guarantee silently degrades into "the orchestrator commits on the
subagent's behalf after inspecting its output," which is a different
(weaker) discipline than the one documented, even though the end state
(a commit exists) looks the same from the log.

This recurred on two back-to-back phases in the same build, not once —
suggesting it isn't incidental to this particular session's tool
configuration, but a structural gap in what delegated phase subagents
are guaranteed to have.

## Suggested fix

Either:
1. Document explicitly (in `hedgehog-landing-loop` and/or the relevant
   agent definitions) that phase subagents require Bash tool access as
   a precondition for delegation, and that the orchestrator should
   verify this before delegating a commit-bearing phase to a subagent;
   or
2. Change the convention so the phase's *content* can be produced by a
   subagent but the commit is explicitly the orchestrating session's
   responsibility to perform immediately upon receiving that output —
   making the current de facto behavior the documented one, rather than
   a silent fallback.

Either resolves the ambiguity; right now the discipline document and
the actual failure mode disagree, and future builds will hit this
identically until one of them changes.

## Source

Build session for a landing-page core project, phases: systems
(commit `082dd8f`), sequence (commit `1c482f7`).

PR #20, opened 2026-08-08
fix(core): fail the commit gate closed, and make the first commit of a fresh install possible

reinaldosaraiva:
Two bugs in the `full-stack-app` core's commit gate, both confirmed against the real files and a real lefthook binary.

## Bug 1 — the gate fails OPEN when lefthook is missing

The payload contains no hook file; hooks are generated by `lefthook install`. Reading the `.git/hooks/pre-commit` that lefthook 2.1.10 (the version this core pins) generates, the lookup chain ends:

```sh
    else
      echo "Can't find lefthook in PATH"
    fi
  fi
}
call_lefthook run "pre-commit" "$@"
```

No `exit 1` — the function returns 0 and the hook exits 0. Reproduced live: with `node_modules` removed, `git commit` printed `Can't find lefthook in PATH` and **exited 0 with the commit landed**.

What makes this dangerous is the appearance. `.git/hooks` is present and `lefthook.yml` comes back with bootstrap, so the repo looks gated when it is not, and nothing in the output distinguishes "passed the gate" from "there was no gate". On the build where this surfaced, five commits after a workspace clean produced ten warnings and zero blocks.

## Bug 2 — the first commit of a fresh install cannot pass

`src/golden-cores/full-stack-app/lefthook.yml` at HEAD:

```yaml
    typecheck:
      glob: '*.{ts,tsx}'
      run: npx nx affected -t typecheck --base=HEAD
```

…identically for `lint` and `test`. In a repo with no commits `HEAD` does not resolve, so all three fail with `fatal: ambiguous argument 'HEAD'` and the commit is rejected. The Nx daemon then dies with `ECONNRESET`, masking the cause. The only way through is `git commit --no-verify`.

`bin/cli.mjs` printed `1. git add -A && git commit -m "chore: install Hedgehog"` with `2. pnpm install` *after* it — so the documented first step could not work.

## What changed, and why each choice

**Base ref.** Each command is now guarded with `if git rev-parse --verify --quiet HEAD; then … --base=HEAD; else npx nx run-many -t <target>; fi`.

`run-many` rather than a skip or an empty-tree SHA, deliberately: with no history there is no diff to narrow by and every project is new, so the whole workspace is both the honest answer and the strongest one — and silencing the gate on the commit that introduces the entire codebase is the worst possible place to do it. The empty-tree-SHA route is also fragile, since Nx's `--head` defaults to `HEAD`, which is equally unresolvable there.

**Fail closed.** `assert_lefthook_installed: true` — lefthook's own switch, which bakes a refusing branch into the generated hooks. Chosen over shipping a shim via `core.hooksPath` because lefthook's postinstall runs `lefthook install -f` on **every** install, so a tracked hook file would be clobbered and would dirty the worktree on each `pnpm install`. Verified: the commit is now refused with exit 1, and the message names `--no-verify` and `LEFTHOOK=0`.

The lockout trade-off seems right because the failure is loud, immediate, self-describing, and repaired by one `pnpm install` — whereas fail-open is silent and surfaces later as unreviewed code already in history.

**Installer ordering.** Swapped to `pnpm install` first, then commit. lefthook's postinstall is what writes the hooks, so the old order guaranteed commit #1 was ungated regardless of the base-ref bug.

I did **not** make the installer create the commit itself. `init` can run in a repo that already has history, and a tool committing into someone's repo unasked is a bigger surprise than a wrong ordering.

**`hedgehog status`.** New `src/db/gate.mjs` (~150 lines), wired into `statusCommand`. It distinguishes four states — never installed / hooks predating the flag (i.e. still fail-open) / fail-closed with the binary gone / active — and names the repair for each. Silent on cores without a `lefthook.yml`, so landing-page is unaffected.

## Evidence

| reproduction | before | after |
|---|---|---|
| `repro/commit-gate-fails-open.mjs` | FAILED (3): commit landed, exit 0, `Can't find lefthook in PATH` ×2 | OK — refused, 0 commits, names `--no-verify` |
| `repro/first-commit-rejected.mjs` | FAILED (3): `fatal: ambiguous argument 'HEAD': unknown revision` ×3, 0 commits | OK — 1 commit, `nx run-many -t {typecheck,lint,test}` all ran |

Each uses its own `mkdtemp` with the git repo one level below it, isolated git config, and an explicit `cwd` — the surrounding repo is never touched. `node scripts/check.mjs` passes.

## Caveats, honestly

- **Real Nx was not run.** `first-commit-rejected.mjs` stubs `npx`, modelling one behaviour: nx resolving `--base` against git. The stub was validated to reproduce the exact real symptom against real lefthook before being relied on. `commit-gate-fails-open.mjs` stubs nothing.
- **The refusal message does not name `pnpm install`.** Lefthook's text is fixed and not configurable — it names `--no-verify` and `LEFTHOOK=0` only. The repair command is surfaced through `hedgehog status` and the `lefthook.yml` comments instead; getting it into the hook itself would require the `core.hooksPath` approach rejected above.
- **The flag is baked in at install time**, so hooks generated before this change stay fail-open until regenerated. That is precisely the state `hedgehog status` now detects and names.
- **Incidental finding, worth knowing:** lefthook's fallback chain tries `pnpm lefthook -h`, and pnpm v11 will install the missing dependency to satisfy it — so on a machine with pnpm and network access, a gateless repo can silently reinstall lefthook mid-commit. The first draft of the fail-open reproduction passed for exactly this wrong reason; both reproductions now pin `PATH` to the binaries the hook actually needs.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -423,8 +423,17 @@ async function init({ force, core, explicitCore, host = DEFAULT_HOST, hostOnly =
   );
   console.log('Next steps:');
   if (explicitCore) {
-    console.log(`  1. ${bold('git add -A && git commit -m "chore: install Hedgehog"')}`);
-    console.log(`  2. ${bold('pnpm install')}`);
+    // `pnpm install` before the first commit, not after it. The core's
+    // commit gate is lefthook, and lefthook's hooks are written by its
+    // own postinstall — so a commit made before the install is a commit
+    // made with no gate at all, and the instruction that put it first
+    // was quietly teaching the project to skip its own discipline on
+    // the one commit that lands the entire workspace. Installing first
+    // means commit #1 is already gated; with no HEAD to diff against it
+    // runs the whole workspace (see lefthook.yml), so expect it to take
+    // as long as a full typecheck/lint/test — that is the gate working.
+    console.log(`  1. ${bold('pnpm install')}`);
+    console.log(`  2. ${bold('git add -A && git commit -m "chore: install Hedgehog"')}`);
     console.log(`  3. Open ${HOSTS[host].label} and describe what you want to build.`);
   } else {
     console.log(`  1. ${bold('git add -A && git commit -m "chore: install Hedgehog"')}`);
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -26,6 +26,7 @@ import { verifyTask } from '../src/db/verify.mjs';
 import { claimTasks, releaseTask, renewLease } from '../src/db/claim.mjs';
 import { readyTasks, formatReady } from '../src/db/ready.mjs';
 import { graphStatus, formatStatus } from '../src/db/status.mjs';
+import { commitGateStatus, formatCommitGate } from '../src/db/gate.mjs';
 import { whyPath, formatWhy } from '../src/db/why.mjs';
 import { addFriction, listFriction } from '../src/db/friction.mjs';
 import { rebuildDb } from '../src/db/rebuild.mjs';
@@ -974,6 +975,18 @@ async function statusCommand() {
   }
 
   console.log(formatStatus(result));
+
+  // Whether the commit gate is actually enforcing. This is reported at
+  // the start of every session because a gate that isn't running looks
+  // exactly like one that is — `.git/hooks` exists, `lefthook.yml` is
+  // committed, and a passing commit prints nothing to tell the two
+  // apart. Nothing else in the discipline would notice.
+  const gate = await commitGateStatus(DEST_ROOT);
+  const gateText = formatCommitGate(gate);
+  if (gateText) {
+    console.log('');
+    console.log(gate.state === 'active' ? dim(gateText) : yellow(gateText));
+  }
 }
 
 // `hedgehog ready` — read-only preview of what a `hedgehog claim` call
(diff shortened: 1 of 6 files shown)


skyf0xx:
Both bugs confirmed (`lefthook.yml:1-12` has no `assert_lefthook_installed`; `cli.mjs:426-427` commits before install runs hooks). Direction is right — closes a fail-open, falls back to more checking not less.

One thing to confirm before merge: `first-commit-rejected.mjs` stubs npx, so the `nx run-many` fallback has never run against real Nx. Worth testing against a fresh generated workspace.

---

skyf0xx:

---

skyf0xx:
Landed on master at 5e5f902, rebased onto current master. One trivial import-list conflict in bin/cli.mjs against #18's boundary import — combined both.

Verified before landing: `node scripts/check.mjs` passes, and the full repro suite passes except the pre-existing, unrelated 05-shipped-cores-unchanged.mjs fixture-staleness failure.

Your two own reproductions (first-commit-rejected.mjs, commit-gate-fails-open.mjs) initially failed on first run here — "lefthook did not install a pre-commit hook; nothing to test" — but that turned out to be a one-off flake in the throwaway npm install, not a real problem: isolating gitEnv/LEFTHOOK/mkdtemp individually all worked, and re-running the actual scripts (3 times) passed cleanly and consistently every time, hook installing and the gate genuinely running (real nx invocations logged) in both the pass and fail-closed cases. Worth knowing about if you see it again, but nothing in this PR's logic is implicated.

Solid diagnosis on both bugs — the fail-open behavior being indistinguishable from a passing gate is exactly the kind of thing that erodes trust in the mechanism silently.

---

PR #23, opened 2026-08-08
fix(verify): judge a task on what it changed, not on everything dirty in the repo

reinaldosaraiva:
`hedgehog verify` judges a task on **everything dirty in the repository**, not on what that task changed. `verifyTask` calls `changedPaths()` with no pathspec, so every dirty path becomes a candidate violation and the whole tree is attributed to whoever verifies first.

On a real build this blocked a task whose agent had touched nothing outside its scope: an uncommitted friction log elsewhere in the repo was enough.

## What I found differs from how I reported it

I opened this expecting a tracked/untracked asymmetry — tracked files blocking, untracked ones silently ignored. **That is wrong, and the experiment says so.** `changedPaths` reads both halves symmetrically, and `splitByScope` treats them identically:

```js
const offending = touched.filter((path) => !inScopeSet.has(path) && !neighborOwned.has(path));
```

Probes against master, with a task scoped to `pkg-a/**`:

| probe | result on master |
|---|---|
| tracked `docs/notes.md` dirty, no neighbour in flight | `scope_violation` |
| **untracked** `other/new-dir/file.txt`, no neighbour in flight | **`scope_violation`** |
| untracked file inside a *neighbour's* declared scope | exit 0, commit held only `pkg-a/src/index.js` |
| file dirtied *during* `verify_command` | exit 0, not committed |

So what actually exonerates a concurrent agent's files is a different rule entirely — `splitByScope`'s `neighborOwned` subtraction via `loadOtherInFlightScopes`. The file that looked "ignored because untracked" was in fact inside its own task's declared scope.

Two further corrections while I was in there:

- **"Re-dirtied during the verify command" does not block on its own.** Gate 1 runs before `runVerifyCommand`. The reported "impossible to win by committing first" is real but one step removed: the *previous* run's write is still dirty at the *next* verify's gate 1. That is what is reproduced and fixed here.
- **On the scheduler premise:** `conflict.mjs`'s `globsOverlap` is segment-prefix based, so two tasks declaring the same shared seam do conflict and are never co-scheduled. The reachable hazard is narrower than "the scheduler authorizes what the gate rejects" — it is a seam in *nobody's* declared scope, which is a genuine violation by whichever agent wrote it, merely misattributed to whichever task verifies first.

The defect that survives all of that is the one worth fixing, and it is the original principle: **a task should be judged on what it changed.**

## The fix

`hedgehog claim` snapshots the dirty working tree as a `path -> sha1` map onto the lease (`tasks.claim_snapshot`), and gate 1 drops any path byte-identical to that snapshot.

- `src/db/claim.mjs` — `pathFingerprint` / `snapshotWorkingTree`, taken **before** `BEGIN`, respecting this file's existing rule that no subprocess runs with a sqlite transaction open. One snapshot per batch; cleared on release and on lease reap.
- `src/db/verify.mjs` — `attributedToTask(changedPaths(), task.claim_snapshot)`.
- `src/db/schema.mjs` — the new column plus `ensureTaskColumns`, an idempotent `ALTER` (`CREATE TABLE IF NOT EXISTS` is a no-op on an existing graph). Verified end-to-end against a DB built from the old schema with a task already `building`: the column is added on the fly, verify falls back to master's behaviour, exit 0.

**The gate is not weakened.** Writing a file changes its bytes, so a real out-of-scope write has a different fingerprint and stays attributed. The only exemption is a task that rewrote byte-identical content over pre-existing dirt — which changed nothing and has nothing to commit. A null snapshot falls back to attribute-everything, so the degraded mode is the strict one. The commit set is untouched; it was already scope-restricted.

**Why the alternatives lost.** An ignore list is a blanket amnesty: a listed path could be written out of scope with impunity, which is exactly the "makes a genuine violation pass" case. Its stated justification — harness writes during verify — turned out to be already handled. It also cannot be implemented within these files, since it would need a `core.yaml` field carried through `core.mjs` and `plan.mjs`, and that is a planner-owned surface and a design decision rather than a bug fix. Treating tracked and untracked differently was moot once the symmetry above was established, and inheriting `.gitignore` adds nothing because `--exclude-standard` already applies.

## Known limitation, deliberately pinned by a test

A path outside every declared scope, changed during the lease by someone else, still blocks whoever verifies first. Exonerating that would make any out-of-scope write forgivable whenever another task happens to be in flight. `repro/05` asserts it keeps failing, so a future change cannot loosen it by accident.

## Evidence

`node --experimental-sqlite repro/run-all.mjs` — a shared `lib.mjs` plus five scripts, plain assertions, no framework, `node:sqlite`, driving the real CLI in throwaway temp dirs.

| | before | after |
|---|---|---|
| 01 genuine out-of-scope write (tracked, untracked, and re-dirtying a pre-existing file) | pass | pass |
| 02 unrelated dirty tracked file, plus a shared seam under a real two-task fan-out | **9 assertions fail** | pass |
| 03 unrelated untracked file, plus contrast cases | **3 fail** | pass |
| 04 telemetry file dirty at claim and re-dirtied during verify | **3 fail** | pass |
| 05 known limitation above | pass | pass |

`run-all` exits 1 before and 0 after. Script 01 passing on **both** sides is the load-bearing one — it is the regression guard proving the gate still catches a real violation. `node scripts/check.mjs` passes.

## Conflict with #16

Expected and textual only. The hunks here touch the file header, the import line, a new `attributedToTask` above `loadOtherInFlightScopes`, `setTaskStatus`'s UPDATE, `claimForVerify`'s opening, and the `withCommitLock` block. #16 rewrites the internals of `changedPaths`/`commitTouchedPaths`/`classifyArtifacts` and deletes `quotePathspec`.

**No new uses of `quotePathspec` are added here**, and the new git calls in `claim.mjs` already use `execFileSync` with argv arrays — so the two changes are compatible in substance and only collide in text. Happy to rebase onto whichever lands first.

### DIFF
--- a/repro/01-genuine-violation-still-fails.mjs
+++ b/repro/01-genuine-violation-still-fails.mjs
@@ -1,0 +1,82 @@
+#!/usr/bin/env node --experimental-sqlite
+// REGRESSION GUARD — passes both before and after the fix, on purpose.
+//
+// The whole point of the scope gate is to catch a task writing outside
+// its ALLOWED SCOPE. Narrowing attribution must not narrow that away, so
+// this asserts the case that must keep failing: the task's own agent
+// writes a file outside scope during the lease, and verify blocks it.
+//
+// Both flavours are covered — modifying a tracked file out of scope, and
+// creating an untracked one — since the fix touches how both are
+// attributed.
+
+import { makeRepo, cleanup, seedTask, claimOrThrow, runVerify, write, append, taskRow, check, finish } from './lib.mjs';
+
+const NAME = '01-genuine-violation-still-fails';
+console.log(`${NAME}\n`);
+
+// ── a tracked file modified out of scope ──────────────────────────────
+{
+  const dir = makeRepo();
+  try {
+    seedTask(dir, { id: 't1', scope: ['pkg-a/**'] });
+    claimOrThrow(dir);
+
+    // The agent's own work: in scope, plus one file it had no business in.
+    append(dir, 'pkg-a/src/index.js', 'export const x = 2;\n');
+    append(dir, 'docs/notes.md', 'the agent wrote here, out of scope\n');
+
+    const res = runVerify(dir, 't1');
+    const row = taskRow(dir, 't1');
+    check('tracked out-of-scope write: verify exits non-zero', res.code !== 0, true);
+    check('tracked out-of-scope write: reports the offending path', res.out.includes('docs/notes.md'), true);
+    check('tracked out-of-scope write: task blocked', [row.status, row.blocked_reason], ['blocked', 'scope_violation']);
+  } finally {
+    cleanup(dir);
+  }
+}
+
+// ── an untracked file created out of scope ────────────────────────────
+{
+  const dir = makeRepo();
+  try {
+    seedTask(dir, { id: 't1', scope: ['pkg-a/**'] });
+    claimOrThrow(dir);
+
+    append(dir, 'pkg-a/src/index.js', 'export const x = 2;\n');
+    write(dir, 'pkg-b/src/sneaky.js', 'the agent created this, out of scope\n');
+
+    const res = runVerify(dir, 't1');
+    const row = taskRow(dir, 't1');
+    check('untracked out-of-scope write: verify exits non-zero', res.code !== 0, true);
+    check('untracked out-of-scope write: reports the offending path', res.out.includes('pkg-b/src/sneaky.js'), true);
+    check('untracked out-of-scope write: task blocked', [row.status, row.blocked_reason], ['blocked', 'scope_violation']);
+  } finally {
+    cleanup(dir);
+  }
+}
+
+// ── a file dirty *before* the claim and changed *again* by the agent ───
+// The narrowing must not become an amnesty: pre-existing dirt is exempt
+// only while the task leaves it alone.
+{
+  const dir = makeRepo();
+  try {
+    append(dir, 'docs/notes.md', 'pre-existing dirt\n');
+    seedTask(dir, { id: 't1', scope: ['pkg-a/**'] });
+    claimOrThrow(dir);
+
+    append(dir, 'pkg-a/src/index.js', 'export const x = 2;\n');
+    append(dir, 'docs/notes.md', 'and then the agent piled on\n');
+
+    const res = runVerify(dir, 't1');
+    const row = taskRow(dir, 't1');
+    check('further-dirtied pre-existing file: verify exits non-zero', res.code !== 0, true);
+    check('further-dirtied pre-existing file: reports the offending path', res.out.includes('docs/notes.md'), true);
+    check('further-dirtied pre-existing file: task blocked', [row.status, row.blocked_reason], ['blocked', 'scope_violation']);
+  } finally {
+    cleanup(dir);
+  }
+}
+
+finish(NAME);
--- a/repro/02-unrelated-dirty-tracked-file.mjs
+++ b/repro/02-unrelated-dirty-tracked-file.mjs
@@ -1,0 +1,81 @@
+#!/usr/bin/env node --experimental-sqlite
+// THE BUG — fails against master, passes after the fix.
+//
+// A tracked file somewhere else in the repo is already uncommitted when
+// the task is claimed: an unfiled friction log, the user's own
+// half-finished edit, whatever. The task's agent never touches it. The
+// gate reads the whole dirty working tree, so it attributes that file to
+// the task and blocks it for work that is not its own.
+//
+// Two shapes: an ordinary unrelated file, and a tracked shared
+// composition seam (a root DI module / package barrel) left dirty by the
+// previous session — the file the reported fan-out failure was about,
+// here dirtied by someone other than the task being verified.
+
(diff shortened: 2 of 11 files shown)


skyf0xx:
Real bug, confirmed at `verify.mjs:334` — `changedPaths()` runs with no pathspec, so the whole dirty tree becomes candidate `offending`.

Needs a rebase: this predates 7450c84 (verify.mjs rewrite), reports MERGEABLE but the merged result hasn't been judged.

One gap: `dirtyPaths()` fingerprints by content sha1 only, so a file dirtied-then-restored to identical bytes is invisible. Worth a note or a fix.

---

reinaldosaraiva:
Update: an independent automated review of this branch found a hole my own fix opened, and it reproduces in three shapes.

**Metadata changes bypass attribution.** My reasoning was "writing a file changes its bytes, so a real out-of-scope write has a different fingerprint and stays attributed". True of content, false of the two things git records that are not content.

Task scope `pkg-a/**`, an out-of-scope path already dirty at claim time, then a metadata-only change during the lease:

| case | `git diff` says | verify |
|---|---|---|
| `chmod 755 docs/notes.md` | `mode change 100644 => 100755` | **exit 0, complete** — change survives uncommitted |
| symlink retargeted to a same-content target | dirty | **exit 0, complete** |
| tracked file replaced by a symlink to same-content file | dirty | **exit 0, complete** |

Git listed all three as dirty, so `attributedToTask` did see them — and dropped them, because `pathFingerprint` returned the claim-time byte hash. The third is the worst: `readFileSync` follows symlinks, so it was fingerprinting whatever the link pointed at rather than the link.

**Fix:** `lstat` first, then fingerprint what git would commit and nothing it would not.

- an **entry-type prefix** (`f…`/`l:`/`o:`), so a file/symlink swap can never fingerprint the same;
- **symlinks read with `readlinkSync`, never followed** — git stores a symlink as a blob holding its target, so a retarget is a content change;
- for regular files, **the owner-execute bit alone** (`mode & 0o100`, git's 100644 vs 100755) hashed alongside the bytes. Deliberately only that bit: folding in the rest of the permission bits would make the gate fire on changes git itself never records, which would re-break the unrelated-dirty-file exoneration this PR exists for.

Deleted-then-recreated identically fingerprints the same — correct, because git reports that path clean and it never reaches the gate.

Cost is one `lstat` plus one read (or `readlink`) per dirty path at claim time; no subprocess, same order of magnitude as before.

The regression table is the part worth reading:

| | upstream master | content-only fingerprint | with this fix |
|---|---|---|---|
| 01 genuine out-of-scope write | pass | pass | pass |
| 02 unrelated dirty tracked file + fan-out | fail (9) | pass | pass |
| 03 unrelated untracked file | fail (3) | pass | pass |
| 04 dirtied during verify | fail (3) | pass | pass |
| 05 pinned known limitation | pass | pass | pass |
| **06 metadata-only change** | **pass** | **fail (9)** | **pass** |

Row 06 passing on master is the point: master blocked these because it blamed the whole dirty tree, my first fix opened the hole, and this closes it without giving back 02/03/04. `scripts/check.mjs` green; old-schema upgrade path still falls back to attribute-everything.

skyf0xx:
This branch is based on `2bcc5c4`, before `7450c84` fixed a shell-injection path in `src/db/verify.mjs` (string-interpolated `execSync` → argv-based `execFileSync`). This PR also touches `verify.mjs`, so please rebase onto current `master` and re-verify your changes still go through the argv-based `git()` helper rather than reintroducing shell-string interpolation. GitHub will show a clean merge regardless, so this needs a deliberate check, not just conflict resolution.

---

skyf0xx:
Landed on master at 1b14934, rebased onto current master. Good self-correction chain in this PR — the "expecting a tracked/untracked asymmetry, found the real defect was different" writeup and the follow-up commit fixing the fingerprint to cover git's type/mode bits (not just content bytes) both held up under review.

Two mechanical conflicts, same pattern as other PRs in this queue: repro/lib.mjs collided with #12's identically-named file (renamed this PR's harness to repro/scope-gate-lib.mjs, repointed its 6 scripts), and src/db/claim.mjs's import block collided with #12's incompleteDependencies import (kept both).

One real cross-PR bug, not a merge-mechanics issue, caught by running the suite: the second commit here (fix(claim), the git-semantics fingerprint upgrade) changed claimOne's signature to take a claim_snapshot parameter, and claimTasks (the batch path, which this PR does touch) was updated to match — but claimTask (the single-task targeted-claim path, which only exists because #12 added it after this branch was written) still called claimOne with the old 3-argument shape. That's not a hunk anyone could see conflict, since #23's diff never touches claimTask at all — it silently passed the task id into the claim_snapshot column position and left the real 4th placeholder undefined, so the UPDATE's WHERE clause never matched and every targeted claim misreported as 'a concurrent claim took it first'. Surfaced as 3 of #12's own reproductions failing (01-retry-after-failed-verify, 02-retry-after-scope-violation, 04-claim-specific-task) despite this PR never touching those files' logic. Fixed by giving claimTask its own pre-transaction snapshot, matching claimTasks's existing pattern, and folded into the fix(verify) commit since that's where claimOne's signature actually changed.

Verified before landing: `node scripts/check.mjs` passes, the full 20-script numbered suite passes except the pre-existing, unrelated 05-shipped-cores-unchanged.mjs fixture-staleness failure, and all 6 of this PR's own reproductions pass individually (including 06-metadata-only-change, the chmod/symlink case from your follow-up commit).

---

PR #24, opened 2026-08-08
feat(cli): stop plan from spawning a server nobody asked for, and let a layer declare the binaries it needs

reinaldosaraiva:
Two independent papercuts, both of which bite hardest in the mode Hedgehog is designed for: an agent running headless.

## Bug 1 — `plan` always starts a graph server, with no way to say no

`planCommand()` took no arguments at all — the dispatcher called it with none, so there was no flag surface:

```js
if (result.compiled.length > 0) {
  const { port } = await startOrReuseGraphServer();
  openInBrowser(`http://localhost:${port}`);
}
```

`startOrReuseGraphServer()` spawns `src/db/graph-server.mjs` detached and unref'd; the child writes `{pid, port}` to `.hedgehog/graph-server.json` on listen and removes it only on SIGINT/SIGTERM. Reproduced before touching anything: one `plan` left pid 2604269 listening and the pidfile on disk. `graphCommand(args)` has had `--no-open` all along.

**Fix, and the default I chose:** the server is now **opt-in via `--open`**. `plan` with no flag starts nothing — no server, no pidfile, no orphan process.

The reasoning for not simply mirroring `graph`'s flag: `graph`'s whole purpose *is* the server, so its `--no-open` only suppresses the browser. On `plan` the server exists solely to be opened, so suppressing the open leaves nothing worth serving. And since every agent-driven run is headless, the case where the server is useful is the exception — it should be the one that has to say so. `--no-open` is still accepted (same no-server behaviour); `--open` together with `--no-open` is a usage error.

Also added display detection (`hasDisplay()` → `presentGraphUrl()`): with no `DISPLAY`/`WAYLAND_DISPLAY` on Linux, both `plan --open` and `graph` print the URL rather than firing `xdg-open` into the void.

## Bug 2 — verify commands inherit an undeclared PATH

A layer's `verify` runs through `/bin/sh -c`, inheriting the caller's environment. On a real build `terraform` lived in `~/.local/bin`, which the user's profile adds to interactive login shells but not to non-interactive ones. The verify passed when a human ran it from their terminal and failed with an opaque **exit 127** when an agent ran it — precisely the mode the tool is built for.

One correction to how I described this: it is not that "nothing validates required binaries", it is that **there was nothing to validate**. `parseCoreYaml` only ever built `{id, depends_on, scope, verify, commit, exclusive, verify_radius}`, `validateCore` checked `id`/`scope`/`verify` plus the module-axis rule, and `status.mjs` never saw the core at all — `graphStatus(db)` and `formatStatus()` are purely graph-derived.

**Fix:** a new optional `requires: [...]` on a layer, plus `src/db/requires.mjs` to check it. The lookup is a PATH walk via `node:fs` rather than a `which` subprocess, because the question is what the verify command's own shell would find (PATHEXT handled for win32). `requires` defaults to `[]`; `validateCore` rejects only a *malformed* declaration, never an absent one, so every existing `core.yaml` stays valid.

Two consumers:

- `hedgehog status` prints a `MISSING BINARIES` block **above** `READY` — above, because a missing binary invalidates every task listed below it.
- `hedgehog verify` gates in `bin/cli.mjs` before calling `verifyTask`. It changes no state: the task keeps its lease and its `building` status, so installing the binary and re-running is the whole fix, and no failed verification is recorded for a command that never ran.

Live, with `requires: ["opentofu"]`:

```
MISSING BINARIES
  opentofu   required by layer "infra" (declared in core.yaml requires:)
...
Missing required binaries. Task CLUSTER-INFRA (layer infra) was not verified.
Not found on PATH: opentofu
```

against the previous `Verification failed. Task ALPHA-INFRA is now blocked (exit 127).`

The field is documented in `hedgehog-core-design`, since a core author would otherwise never learn it exists.

## Evidence

`repro/run-all.mjs` — three scripts plus a shared `_lib.mjs`, standalone Node, no framework. Temp dirs use a `hedgehog-papercuts-` prefix, and cleanup kills graph-server processes matched by *that project's own path in argv*, never by a name glob.

- **Before** (fixes stashed), `run-all` exits 1: 5 failures in repro 1 (pidfile present and process alive after both `plan --no-open` and bare `plan`), 6 in repro 2 (no `MISSING BINARIES`, verify reported exit 127 and left the task blocked), 2 in repro 3 (`requires` was `undefined` on all 8 full-stack-app layers and all 5 landing-page layers).
- **After:** all 36 checks pass, with no temp dirs and no graph-server processes left behind.

`node scripts/check.mjs` passes. Both shipped cores were dumped through the loader before and after and diffed with `requires` stripped: **identical**, with `requires` an empty list on every layer of both.

## Not done, and why

- **`next`/`claim` packets do not surface a layer's `requires`.** The declaration reaches the agent only through `status` and the verify gate. Surfacing it would need a new `tasks` column and a schema migration, which seemed out of proportion to the papercut.
- **The verify-time gate lives in `bin/cli.mjs`, not `verify.mjs`,** because a change to that file is already open upstream (#16) and I was asked not to touch it. So the gate covers the CLI path only; a direct `verifyTask()` caller still gets the raw 127. Closing that is roughly ten lines in `verify.mjs`: `runVerifyCommand` taking the layer's `requires`, checking before `execSync`, and returning a distinct outcome (e.g. `missing_binaries`) so the task is not marked `verification_failed` — and `verifyTask` would need the core passed in, since the task row does not carry `requires`.
- `plan --open` still only opens when at least one intent compiled, matching the previous behaviour.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -797,6 +797,22 @@ async function nextCommand() {
   console.log(formatNext(packet));
 }
 
+// Mirrors, read-only, the condition verifyTask's own Phase 0
+// (claimForVerify) enforces: after expired leases are reaped, the task
+// must be `building` and leased to `owner`. The expiry predicate is
+// claim.mjs#reapExpiredLeases's, kept identical so this can't disagree
+// with the check it's standing in for. Writes nothing and reaps nothing
+// — the authoritative check, with its state changes, stays in verify.mjs.
+const VERIFIABLE_BY_SQL = `
+  SELECT 1 FROM tasks
+  WHERE id = ? AND status = 'building' AND lease_owner = ?
+    AND (lease_expires_at IS NULL OR lease_expires_at >= datetime('now'))
+`;
+
+function taskVerifiableBy(db, taskId, owner) {
+  return db.prepare(VERIFIABLE_BY_SQL).get(taskId, owner) !== undefined;
+}
+
 // The declared-binary gate for one task: resolves the task's layer in
 // the core definition and returns { layer, missing } when that layer's
 // `requires:` names binaries this environment can't find, or null when
@@ -808,7 +824,17 @@ async function nextCommand() {
 // so installing the binary and re-running `hedgehog verify` is the whole
 // fix — no unblocking step, and no failed verification recorded for
 // something that never ran.
-async function missingBinariesForTask(db, taskId) {
+//
+// It only speaks for a caller who could actually verify this task. A
+// caller holding no lease, or a stale one, has a different problem, and
+// telling them to install a binary would send them off to fix something
+// that isn't in their way — so this stays quiet and lets verifyTask
+// report the real ownership/state error. Nothing is lost by deferring:
+// claimForVerify throws before any verify command runs, so falling
+// through still records no failed verification.
+async function missingBinariesForTask(db, taskId, owner) {
+  if (!taskVerifiableBy(db, taskId, owner)) return null;
+
   const row = db.prepare('SELECT layer FROM tasks WHERE id = ?').get(taskId);
   if (row === undefined) return null;
 
@@ -852,7 +878,7 @@ async function verifyCommand(args) {
   const db = openDb();
   let result;
   try {
-    const blockers = await missingBinariesForTask(db, taskId);
+    const blockers = await missingBinariesForTask(db, taskId, owner);
     if (blockers) {
       const list = blockers.missing.map((b) => bold(b)).join(', ');
       console.error(
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -305,8 +305,9 @@ ${bold('Usage')}
   npx @skyf0xx/hedgehog update                    refresh the installed agents + skills
   npx @skyf0xx/hedgehog db init                   create .hedgehog/hedgehog.db if absent
   npx @skyf0xx/hedgehog db rebuild                re-derive the build graph from committed intents + git history
-  npx @skyf0xx/hedgehog plan                      compile pending intents into tasks + dependencies,
-                                                   then open the build graph if anything compiled
+  npx @skyf0xx/hedgehog plan                      compile pending intents into tasks + dependencies
+                                                   (starts no graph server; --no-open says so explicitly)
+  npx @skyf0xx/hedgehog plan --open               also start the graph server and open it, if anything compiled
   npx @skyf0xx/hedgehog intent add [flags]        add an intent (rules/requirements/dependencies)
   npx @skyf0xx/hedgehog intent add --file <path>  add an intent from a JSON file
   npx @skyf0xx/hedgehog next                      print the task packet for one ready task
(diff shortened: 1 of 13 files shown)


skyf0xx:
Clean, `requires:` gate is additive only. One thing needing your call: `plan` no longer starts a graph server by default (opt-in via `--open`) — UX default flip, not a bug fix.

Collides directly with #29 in `core.mjs`'"'"'s `parseCoreYaml` layer literal — whoever lands second is a hand-merge.

---

reinaldosaraiva:
Update: an independent automated review raised two on this branch. Both hold, and one came with a suggested patch I did **not** take verbatim — the experiment that confirmed the finding also turned up a case the patch gets wrong.

**1 — Empty PATH entries resolve incorrectly.** Confirmed the POSIX premise against the real shell before accepting anything. With a probe executable in the cwd, both `/bin/sh` (dash) and `/bin/bash` ran it for all four shapes — leading, trailing, interior, and `PATH=""`:

```
PATH=":/usr/bin"        shell=FOUND     findBinary=null
PATH="/usr/bin:"        shell=FOUND     findBinary=null
PATH="/usr/bin::/bin"   shell=FOUND     findBinary=null
PATH=""                 shell=FOUND     findBinary=null
```

So the false negative is real, in the build-blocking direction.

The suggested `dir || '.'` is not quite right, though: **an unset PATH is not an empty one.** `env -u PATH sh -c 'probe'` fails with 127 while `ls` still resolves — the shell falls back to confstr `_CS_PATH` (`getconf PATH` → `/bin:/usr/bin`) and does *not* search the cwd. A blanket `dir || '.'` would turn unset-PATH into a cwd search, which is a small security-shaped difference rather than a cosmetic one.

Fixed with a `searchPath(env)` helper: an empty component becomes `.` only when PATH is actually set; unset falls back to `/bin:/usr/bin`. Windows folds in from the other side — cmd.exe ignores empty entries but always searches the cwd first, so `.` is prepended there. The POSIX behaviour is verified by execution; **the Windows branch is reasoned, not executed** — no Windows host here.

**2 — Prerequisite check bypasses lease validation.** Reproduced: a task leased to alice, bob runs verify with a `requires:` naming a missing binary → bob is told to install it. Control on a core without `requires:` → `not leased to bob (currently: building, leased to alice)`.

Reordered rather than reworded, and the reordering is safe: `verifyTask` → `claimForVerify` is Phase 0, throwing on the lease/status check *before* the scope gate and before `runVerifyCommand`. I asserted the `verifications` table is empty in all four cases rather than inferring it. The gate now short-circuits only when a read-only mirror of `claimForVerify`'s condition holds — status `building`, leased to this owner, lease unexpired using that function's own predicate, kept identical so the two cannot drift. The mirror writes nothing and reaps nothing. So the no-failed-verification property is preserved *and* the caller gets the honest error, with no need for a hedged message.

**Incidental finding, not fixed** (it is `src/db/verify.mjs`, which I was asked to leave alone, and it predates this branch): on the expired-lease path the message and the database disagree. `claimForVerify` reaps inside its transaction, reads the freshly-reaped row so the caller sees `currently: blocked`, then throws — and the `ROLLBACK` in its catch undoes the reap, leaving the row `building`/alice. Observed directly. The reproduction asserts it as-is, with a comment, rather than encoding what it arguably should be.

Reproductions: `path-empty-entry.mjs` 6 of 12 failing before, all passing now — it proves the shell's behaviour first and only then asserts `findBinary` matches, so the premise is tested rather than assumed. `verify-lease-precedence.mjs` 7 of 21 failing before. Full suite green, `scripts/check.mjs` green.

skyf0xx:

---

skyf0xx:
Landed on master at 3f02446, rebased onto current master. Both bugs merged together — the requires:/PATH-checking fix (Bug 2) is a clean, backward-compatible addition, and the plan default-flip to opt-in server (Bug 1) is the right call for headless/agent-driven use, the tool's primary mode, with --open preserving the old behavior for anyone who wants it.

Many small conflicts, all mechanical: bin/cli.mjs's imports and planCommand's two additions (--recompile from #17, --open/--no-open from this PR) coexist cleanly with recompile short-circuiting first; src/db/core.mjs's per-layer validation loop combined this PR's requires check alongside #28's/#29's existing ones; src/db/status.mjs's formatStatus gained both missingRequirements and drift as parameters; repro/_lib.mjs and repro/run-all.mjs collided with #29's identically-named files (same pattern as every other PR in this queue) — renamed this PR's harness to papercuts-lib.mjs. Also caught and removed a genuine duplication the merge introduced: statusCommand was resolving and loading core.yaml twice (once for drift, again for missingRequirements) — simplified to reuse the one already loaded.

One real bug found by actually running this PR's own core-without-requires.mjs reproduction, not caused by this PR: validateCore/lintCore's verify_radius checks tested `!== null`/`=== null` only. Every core.yaml-sourced core is fine since the parser always produces null for an unset field, but this PR's own test constructs a core object directly to test validateCore in isolation, leaving verify_radius genuinely undefined — and undefined !== null crashed on .length. Latent since #28 introduced these checks; this PR's test was just the first thing to exercise that path. Fixed both checks to tolerate undefined alongside null, in a separate small commit.

Two things left as-is, not fixed: the pre-existing, unrelated 05-shipped-cores-unchanged.mjs fixture-staleness failure (tracked separately across this whole session), and one assertion in plan-no-open.mjs ("prints the URL when there is no display") that's inherently Linux-only — hasDisplay() correctly returns true unconditionally on darwin/win32 per your own documented design, so testing it with DISPLAY='' on this macOS machine exercises a branch that can't apply here. The other 10 assertions in that script, and all of requires-missing-binary/core-without-requires/path-empty-entry/verify-lease-precedence, pass clean.

Your own two self-correction commits (empty-PATH-component handling confirmed against real dash/bash behavior, and the lease-error-outranks-missing-binary ordering fix) both held up under review — good instinct catching the second one, since a wrong-owner or expired-lease caller being told to install a binary they didn't need is a genuinely confusing failure mode.

---

PR #25, opened 2026-08-08
fix(full-stack-app): repair verify commands that cannot run, and lint the core definition so they cannot recur

reinaldosaraiva:
Every project created from the `full-stack-app` Golden Core starts with verify commands that cannot run. This was found while checking an authoring rule against reality, and validated against a **real instantiated workspace** rather than by reading.

## Three confirmed defects

**1. `--testPathPattern` is a Jest flag; this core runs Vitest 4.** Six occurrences across five lines (`schema`, `contract`, `controller`, `hook`, and `screen` which has two). `package.json` pins `"vitest": "~4.1.0"`. Baseline against the real workspace:

```
> nx run db:test --testPathPattern=board
> vitest --testPathPattern=board
CACError: Unknown option `--testPathPattern`   → exit 1
```

The gate aborts before running a single test.

**2. `pnpm typecheck` does not exist.** The core's `package.json` has exactly one script, `dev`. The `join` layer's verify begins with `pnpm typecheck`. Baseline: `ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "typecheck" not found`.

**3. An unanchored `{module}` filter — and this one is worse than a wrong match.** A bare `{module}` is a substring match against the whole path, so `card` also selected `src/board/use-board-cards.spec.ts` (measured: 2 files vs 1).

But `list` is a **Vitest subcommand**, alongside `run`, `related` and `watch`. So for a module named `list` the command becomes `vitest list`, which collects and prints tests and exits 0. Measured, with a deliberately broken `use-list.ts` in place:

```
pnpm nx test hooks -- list        → exit 0     (the `list` subcommand; nothing ran)
pnpm nx test hooks -- src/list/   → exit 1     FAIL src/list/use-list.spec.ts
```

A module named `list` gets a gate that **cannot fail**.

## Validated against a real workspace

`node bin/cli.mjs init --ts-full-stack-app` into a temp dir, `pnpm install` (exit 0, 1.2 GB, lockfile byte-identical to the shipped one), then modules `board`/`list`/`card` scaffolded across all eight layers, plus the projects the core does not ship. Docker was unavailable but is not needed — these tests are pure.

## The fix

| layer | verify |
|---|---|
| schema | `pnpm nx test db -- src/schema/{module}/` |
| contract | `pnpm nx test contracts -- src/{module}/` |
| repository / service | unchanged |
| controller | `pnpm nx test api -- src/app/{module}/` |
| hook | `pnpm nx test hooks -- src/{module}/` |
| screen | `pnpm nx test web -- src/app/{module}/` |
| join | `pnpm nx run-many -t typecheck,lint,test` |

`nx test <project> -- <path>` forwards positionally, verified: `> nx run db:test src/schema/board/` → `> vitest src/schema/board/`. The Nx form was kept rather than `pnpm --filter … exec vitest`, because this core's `verify_radius` reasoning is written against `nx test` and `dependsOn: ["^build"]` is part of the gate.

For `join`, `run-many -t typecheck,lint,test` was chosen over adding a `typecheck` script: it is the exact sweep the bootstrap skill already verifies the workspace with, and it reaches the `typecheck` target `@nx/js/typescript` registers, leaving no root script to drift out of sync.

## A fourth defect, which `core.yaml` genuinely cannot express

`screen`'s verify also ran `pnpm nx test mobile`, but `apps/mobile` only exists when the Mobile add-on is on. With it off there is no project and no target.

The tempting workaround was measured and rejected: `nx run-many -t test -p mobile` exits **0** with "No tasks were run" when the project is absent — and equally when it exists with no test target. That is a gate that silently does nothing, which is the failure mode this core keeps hitting.

`core.yaml`'s grammar is flat scalars with no conditional, so it cannot express "only when this add-on is on". So `screen` now verifies web only; the mobile scope glob stays, since the layer must still be allowed to write there; `join`'s sweep covers `apps/mobile`; and extending the verify belongs to the add-on's own bootstrap step. All of this is documented in the file's header rather than left implicit.

## Evidence

All eight commands were materialized straight out of the fixed `core.yaml` and run against the workspace:

- **Pass:** all 8 exit 0 (`join` = typecheck + lint + test across 10 projects).
- **Fail:** each layer's own artifact broken → all 8 exit 1. `join` was given a **mobile-only** break and still caught it, which is what backs the mobile decision above.
- **Anchoring:** broken board file → `-- card` exits 1 (wrongly failing card's gate), `-- src/card/` exits 0. Broken list file → `-- list` exits 0, `-- src/list/` exits 1.
- **Empty filter:** a module with source and no tests → `No test files found, exiting with code 1`, so a testless layer is caught rather than waved through.

The failing half is the load-bearing one. A gate that runs but cannot fail is the same defect in a different costume.

## A lint over the core definition

`repro/core-verify-lint.mjs` checks both shipped cores for: Jest-only flags; `pnpm <script>` references that do not resolve; unanchored `{module}` filters; and `nx test <project>` naming a shipped project with no test target. It ends with the literal 4.0.3 strings as a regression fixture, so it cannot quietly stop detecting what it was written for.

Against the original `core.yaml` it reports all six defects and exits 1; against the fixed one, exit 0. `node scripts/check.mjs` passes.

This lint is arguably worth more than the fix — it would have caught all of this at authoring time.

## One confirmed defect deliberately left open

**`apps/api` and `apps/web` ship no vitest config, so they have no `nx test` target at all.** Baseline: `NX Cannot find configuration for task api:test` / `web:test`. The `controller` and `screen` verifies therefore still cannot run on a freshly bootstrapped core, whatever the filter says.

Corroborated independently: a real project built on this core had to hand-add `apps/api/vitest.config.mts`, `apps/web/vitest.config.mts` + `vitest.setup.ts`, both `tsconfig.spec.json` files, and `jsdom` / `@vitejs/plugin-react` / `@testing-library/*` devDependencies.

Fixing it means new dependencies, a regenerated 734 KB lockfile, and placeholder specs — a vitest config with no test files would make bootstrap's own `nx run-many -t test` fail. That is a core regeneration via `scripts/regenerate-full-stack-app-core.sh`, not a verify-command edit, so it is out of scope here. It is recorded as `KNOWN_MISSING_TEST_TARGET` in the lint, which fails if that list ever goes stale in either direction. The command *form* for those two layers was proved by adding the missing targets in the throwaway fixture.

## Also outside this PR's paths

`src/skills/hedgehog-core-design/SKILL.md` still quotes `--testPathPattern={module}` as exemplary at lines 227 and 249, and `pnpm typecheck` at line 267. Left untouched here to keep this PR to the core definition, but they will re-teach the bug to every future core author.

### DIFF
--- a/repro/core-verify-lint.mjs
+++ b/repro/core-verify-lint.mjs
@@ -1,0 +1,277 @@
+#!/usr/bin/env node
+// Lint over the shipped Golden Cores' verify commands, plus the regression
+// fixture for the bugs it was written from.
+//
+// Run: node repro/core-verify-lint.mjs
+//
+// ── What went wrong ──────────────────────────────────────────────────
+// src/golden-cores/full-stack-app/core.yaml shipped verify commands that
+// could not run at all against the workspace the same directory ships:
+//
+//   1. `pnpm nx test db --testPathPattern={module}` (and five more like it).
+//      `--testPathPattern` is a Jest flag. This core pins `vitest ~4.1.0`,
+//      and Vitest's CLI parser rejects unknown options outright:
+//        CACError: Unknown option `--testPathPattern`
+//      Every one of those gates aborted before running a single test — a
+//      gate that could never pass, whatever the code did.
+//
+//   2. `pnpm typecheck && pnpm nx run-many -t test` on the join layer.
+//      The core's package.json defines exactly one script, `dev`:
+//        ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "typecheck" not found
+//
+//   3. A bare `{module}` as the test filter. Vitest's positional filter is
+//      a substring match against the whole test file path, so a module
+//      named `card` also pulled in `src/board/use-board-cards.spec.ts` —
+//      and a module named `list` hit `vitest list`, Vitest's
+//      list-collected-tests subcommand, which prints the collected tests
+//      and exits 0. That is a gate that cannot fail.
+//
+// Both (1) and (2) are decidable from the core definition alone — no
+// workspace, no install, no network. That is what this file automates, so
+// the next core that ships a verify command naming a flag or a script that
+// does not exist fails here instead of on someone's first build.
+
+import { readFile, readdir, access } from 'node:fs/promises';
+import { join, dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { parseCoreYaml, validateCore } from '../src/db/core.mjs';
+
+const ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '..');
+const CORES_DIR = join(ROOT, 'src/golden-cores');
+
+const failures = [];
+const fail = (msg) => failures.push(msg);
+
+// ── Flags that belong to Jest and do not exist in Vitest's CLI. Vitest
+//    exits on the first unknown option, so any of these is a gate that can
+//    never pass. ────────────────────────────────────────────────────────
+const JEST_ONLY_FLAGS = [
+  'testPathPattern',
+  'testPathPatterns',
+  'testPathIgnorePatterns',
+  'runTestsByPath',
+  'runInBand',
+  'detectOpenHandles',
+  'forceExit',
+  'collectCoverageFrom',
+  'moduleNameMapper',
+  'setupFilesAfterEnv',
+  'testMatch',
+  'roots',
+];
+
+// ── `pnpm <word>` is a package.json script unless <word> is one of pnpm's
+//    own subcommands or a binary from a declared dependency. ────────────
+const PNPM_SUBCOMMANDS = new Set([
+  'add', 'audit', 'bin', 'config', 'create', 'dedupe', 'deploy', 'dlx', 'env',
+  'exec', 'fetch', 'i', 'import', 'init', 'install', 'licenses', 'link', 'list',
+  'ls', 'outdated', 'pack', 'patch', 'prune', 'publish', 'rebuild', 'remove',
+  'root', 'run', 'server', 'setup', 'start', 'store', 'test', 'unlink', 'update',
+  'why', 'why-command',
+]);
+
+// ── Projects the core ships whose `nx test` target does not exist yet.
+//    Each entry is a known-open defect, not a style exemption: the shipped
+//    apps/api and apps/web carry no vitest config, so they have no `test`
+//    target, and a verify naming `nx test api`/`nx test web` dies with
+//    "Cannot find configuration for task api:test" no matter how the
+//    filter is spelled. Fixing it means adding a vitest config, a spec,
+//    and (for web) jsdom + @testing-library to the core — a core
+//    regeneration, tracked separately. The lint fails if an entry here
+//    becomes unnecessary, so the list cannot rot. ───────────────────────
+const KNOWN_MISSING_TEST_TARGET = {
+  'full-stack-app': ['api', 'web'],
+};
+
+const exists = async (p) => {
+  try {
+    await access(p);
+    return true;
+  } catch {
+    return false;
+  }
+};
+
+// Splits a verify command on shell operators into individual commands.
+const splitCommands = (verify) =>
+  verify
+    .split(/&&|\|\||;|(?<!\|)\|(?!\|)/)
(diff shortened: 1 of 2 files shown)


skyf0xx:
The core.yaml half landed on master in a300678 (4.0.5) — six `--testPathPattern=` flags (Jest syntax, core pins Vitest 4.1) and a nonexistent `pnpm typecheck` script, both confirmed independently before fixing.

Not landed: the lint, since it sits in `repro/` unwired to `scripts/check.mjs`. Would be a welcome follow-up PR.

Correction from #16: `repro/` doesn'"'"'t ship to consumers, `files[]` already excludes it.

Closing as landed — thanks, best report in the queue.

---

reinaldosaraiva:
A question about where verification code can live, prompted by an independent review of this branch.

An automated reviewer flagged that `repro/core-verify-lint.mjs` is **not invoked by `npm run check`**, so the defects it detects — Jest-only flags, `pnpm <script>` references that do not resolve, unanchored module filters — can still pass CI and ship unless someone remembers to run it by hand. That is a fair hit, and it is the same shape as the finding this PR fixes: a gate that exists but never runs.

I did not wire it in, because #16's close said `repro/` would ship as a permanent top-level directory in the package tree. But checking that assumption: **`repro/` is outside `package.json`'s `files` array, so the published tarball is byte-identical either way.** The cost is a directory in the repo, not in what users install.

So the question is narrower than I first assumed, and it is yours to answer:

1. Is the objection to a directory in the **repo tree**, or to shipped weight? If the former, would these fit somewhere you would accept — `scripts/`, a `test/` directory, or folded into `scripts/check.mjs` itself?
2. Specifically for this PR: the core-definition lint seems like it wants to be a permanent release gate rather than a reproduction. It reads both shipped cores and fails on Jest-only flags, unresolvable script references, unanchored `{module}` filters, and `nx test <project>` naming a project with no test target. It also carries the literal 4.0.3 strings as a regression fixture so it cannot quietly stop detecting what it was written for. If you want it, I will move it to `scripts/` and wire it into `check.mjs` in this PR.

Happy either way — I would just rather ask than guess, since the same question applies to the reproductions on the other open branches.

reinaldosaraiva:
Answering my own question above, at least for this branch — an independent review made the same point, so I have wired it in rather than leave it hanging.

`repro/core-verify-lint.mjs` → **`scripts/lint-core-verify.mjs`**, restructured to export `lintCoreVerifyCommands(root)` returning failure messages instead of exiting on its own. `scripts/check.mjs` imports it as section 5c and pushes each message through its existing `fail()` list, so failures print in the same `✗` format as every other section. The summary line names it: `ok — 18 agents, 23 skills, 2 cores, verify commands, tarball, CLI`.

That puts it on both CI paths already in the repo — `.github/workflows/check.yml` on every PR, and `publish.yml`'s gate before `npm publish`. It still runs standalone.

Note on placement, which also answers the tarball half of my question: `scripts/` is not in `package.json`'s `files`, so neither `check.mjs` nor this new lint ships — consistent with how `check.mjs` already worked.

**Red/green, by exit code rather than by output:**

```
core.yaml restored to 4.0.3  ->  node scripts/check.mjs  exit=1   (6 ✗ messages)
core.yaml as fixed here      ->  node scripts/check.mjs  exit=0
```

**The regression fixture survives the move, and I proved it rather than asserting it.** Emptying `JEST_ONLY_FLAGS` — simulating someone gutting the check — makes `npm run check` go red with five messages of the form *"regression fixture no longer detected — the 4.0.3 schema verify `pnpm nx test db --testPathPattern={module}` now passes every check"*, rather than passing silently. So the lint cannot quietly stop detecting what it was written for.

**What stayed in `repro/`, and why the distinction matters:** `full-stack-app-verify-workspace.mjs` needs an instantiated workspace and a 1.2 GB install, and it covers exactly the facts a static lint *cannot* decide — that `nx test <project> -- <path>` forwards positionally, that `vitest list` exits 0 without running a test, that a filter matching nothing exits 1. It asserts facts about Nx and Vitest at pinned versions rather than a permanent property of the shipped product, and it would make `npm run check` a multi-minute job.

By that criterion only the core-definition lint belongs in the release path: a core definition is shipped data, and its verify commands are never exercised by building this package — which is exactly why a broken one shipped in the first place.

PR #27, opened 2026-08-08
feat(packet): make runtime honesty a standing requirement instead of something a human remembers to ask for

reinaldosaraiva:
On a real 36-task build, four different layer agents did notably honest things when they hit the limits of their scope: wired a **noisy stub** that throws a named domain error at first use rather than a permissive one; **refused to invent** cascade-delete semantics the requirements had not decided; rendered *"aggregates unavailable"* rather than a fabricated `0`; and **reported instead of writing out of scope** when the scope turned out to be wrong.

That is real defensive behaviour, and it is why several silent failures were caught at all. But it existed because **each packet was hand-written with an explicit honesty requirement in it**. Nothing in Hedgehog asks for it, and a packet generated from `core.yaml` — which is what the tool actually emits — does not.

The evaluation's conclusion was blunt: the quality did not come from the discipline, it came from a human remembering to ask, every time.

## What the packet says today

`formatNext` renders sections that are entirely about *what* and *where*:

```js
lines.push(`TASK  ${task.id}`); lines.push(task.objective);
lines.push('STATUS   READY');
lines.push('INTENT');             // goal, outcome
lines.push('RELEVANT RULES');
lines.push('WHY NOW');
lines.push('BLOCKED DOWNSTREAM');
lines.push('ALLOWED SCOPE');
lines.push('VERIFICATION');
```

Nothing about *how* to build. `src/agents/layer-eng.md`'s Constraints cover scope, self-certification, and never weakening a verify command — but never the case where the layer **cannot do what it was asked**.

## What this adds

A `HONESTY` section, appended after `VERIFICATION`:

```
HONESTY
  Build what this layer can; make what it can't obvious.
  - A stub or placeholder throws a named error at first use — never
    returns empty, null, or success from something unbuilt.
  - A value that cannot be computed is surfaced as unavailable — never
    replaced by 0, "", or a plausible default.
  - A decision RELEVANT RULES doesn't make is reported, not invented.
  - A scope that turns out to be wrong is reported, not widened.
  Reporting one of these is a successful outcome. Papering over it is
  the failure VERIFICATION cannot catch.
```

Ten lines, because a packet is read on every task and a long section gets skimmed. Each bullet names one of the four observed behaviours, and it refers to sections the packet already carries rather than inventing vocabulary.

The last two lines are the load-bearing ones: an agent that believes reporting a gap counts as failing its task will fabricate instead. Removing that incentive is most of the work.

## Where it lives, and why both places

- **The packet (primary).** It is the only surface guaranteed to be read on every task, it is regenerated per task rather than remembered, and it is emitted by the CLI — so a project running `npx @skyf0xx/hedgehog` gets it on the next version bump, with no `hedgehog update` and no re-install.

  Placing it **last** is deliberate: it qualifies the gate directly above it, so it reads as the answer to "what if I can't clear VERIFICATION honestly" rather than as preamble.
- **The agent payload (reinforcement).** `layer-eng.md`, `backend-eng.md` and `front-end-eng.md` each gain one Constraints bullet, tailored to what that role actually fakes — a permissive stub; a schema semantic like cascade-on-delete that the rules never decided; a screen rendering `0` or placeholder rows for a figure the hook cannot supply. This reaches existing projects only on `hedgehog update`, which is exactly why it cannot be the only home — but it puts the rule where that agent's other behavioural rules live.
- `hedgehog-authored-loop/SKILL.md`, the one place enumerating the packet's sections exhaustively, now lists `…/VERIFICATION/HONESTY`. The abbreviated enumerations in `templates/CLAUDE.md`, `hosts/routing.mjs` and `hedgehog-landing-loop` were left alone: they already elide `INTENT`/`RELEVANT RULES`, so they do not claim exhaustiveness, and touching them would only widen the conflict surface with #19.

Nothing existing was reworded. The only non-additive hunk is a six-line re-wrap in `hedgehog-authored-loop/SKILL.md` — same words, moved across line breaks to keep the paragraph under the file's wrap after the longer section list.

## Evidence

Two reproductions, each building a two-layer authored core in a temp dir and driving the real CLI with `NO_COLOR=1`. (Graph construction goes through the same modules the CLI imports rather than through `hedgehog plan`, because `plan` spawns a detached graph server and an OS browser-open on success.)

**Before**, both exit 1:

```
FAIL  hedgehog next emits an HONESTY section
FAIL  packet states the "noisy stub" case
FAIL  packet states the "no fabricated value" case
FAIL  packet states the "no invented decision" case
FAIL  packet states the "no widened scope" case
--
PASS  the pre-existing packet is unchanged and still leads the output
FAIL  packet sections appear once each, in order
      expected: … ALLOWED SCOPE | VERIFICATION | HONESTY
      actual:   … ALLOWED SCOPE | VERIFICATION
```

**After**, both exit 0, including the ordering assertion across all eight sections.

Packet diff on an unchanged fixture — additions only, at the end:

```diff
@@ -24,3 +24,14 @@
 VERIFICATION
   node --version
+
+HONESTY
+  Build what this layer can; make what it can't obvious.
+  ...
```

`node scripts/check.mjs` passes. `repro/` is outside `package.json`'s `files`, so the published tarball is unchanged.

## What this does and does not buy

It is a prompt-level instruction, and `hedgehog verify` cannot check it. Nothing here detects a permissive stub, a fabricated `0`, or an invented cascade semantic — verify still only asks whether the command exited 0 and whether the writes stayed in scope.

What changes is that the requirement is **standing** rather than **remembered**: it arrives on every task from the tool, instead of depending on a human writing it into each dispatch. That is precisely the failure the evaluation identified, and the behaviour it asks for was already demonstrably achievable by these agents when asked.

Whether they comply unprompted is a question this makes measurable, not one it settles. No mechanical check was attempted, because I do not think a credible one exists at this layer — and a check that looks like enforcement without being it would be worse than none.

## Conflicts

Expected, in different places, with two open PRs:

- **#12** (`claim` emits the full packet) — no conflict in `next.mjs`. If #12 routes `claimCommand` through `formatNext`, HONESTY arrives in claim output for free. A conflict arises only if #12 duplicates the renderer rather than reusing it.
- **#19** (`GOAL`/`OUTCOME` labels + `INHERITED DEBT`) — conflicts in `formatNext`'s body and header comment, and in the skill's section enumeration. Mechanical to resolve: #19 edits and inserts among the mid-packet sections, this only appends after the last `lines.push`. The concerns are orthogonal — theirs is what the intent asked for, this is how to build when you cannot deliver it.

### DIFF
--- a/repro/fixture.mjs
+++ b/repro/fixture.mjs
@@ -1,0 +1,107 @@
+// Shared fixture for the packet reproductions. Builds a throwaway
+// Hedgehog project in a temp dir (its own `hedgehog-honesty-` prefix, so
+// it never collides with another working copy's temp dirs), compiles one
+// intent through a two-layer authored core, and runs the real CLI's
+// `hedgehog next` against it.
+//
+// The graph is built through the same modules the CLI itself imports
+// (dbInit / addIntent / loadCore / planTasks) rather than by shelling out
+// to `hedgehog plan`, because `plan` spawns a detached graph server and
+// an OS browser-open on success — neither belongs in a reproduction. The
+// assertion target, `hedgehog next`, is driven as the real CLI process.
+
+import { mkdtemp, mkdir, writeFile, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join, dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { execFileSync } from 'node:child_process';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+export const REPO_ROOT = resolve(__dirname, '..');
+export const CLI = join(REPO_ROOT, 'bin/cli.mjs');
+
+const CORE_YAML = `id: honesty-fixture
+layers:
+  - id: schema
+    scope: [src/schema/**]
+    verify: node --version
+    commit: "feat(schema): fixture"
+  - id: service
+    depends_on: schema
+    scope: [src/service/**]
+    verify: node --version
+    commit: "feat(service): fixture"
+`;
+
+const INTENT = {
+  id: 'billing',
+  goal: 'Charge a customer for a subscription period',
+  outcome: 'An invoice exists for every closed period',
+  priority: 1,
+  rules: [
+    'An invoice is immutable once issued',
+    'A period with no usage still produces an invoice',
+  ],
+};
+
+// Creates the temp project and returns its absolute path. Caller is
+// responsible for cleanup via `cleanup(dir)` — never by globbing the
+// temp root, which would delete sibling fixtures.
+export async function makeFixture() {
+  const dir = await mkdtemp(join(tmpdir(), 'hedgehog-honesty-'));
+  await mkdir(join(dir, '.hedgehog'), { recursive: true });
+  await writeFile(join(dir, '.hedgehog/core.yaml'), CORE_YAML);
+
+  const cwd = process.cwd();
+  process.chdir(dir);
+  try {
+    const { dbInit, openDb } = await import(join(REPO_ROOT, 'src/db/init.mjs'));
+    const { addIntent } = await import(join(REPO_ROOT, 'src/db/intent.mjs'));
+    const { loadCore } = await import(join(REPO_ROOT, 'src/db/core.mjs'));
+    const { planTasks } = await import(join(REPO_ROOT, 'src/db/plan.mjs'));
+
+    await dbInit();
+    const core = await loadCore(join(dir, '.hedgehog/core.yaml'));
+    const db = openDb();
+    try {
+      await addIntent(db, INTENT);
+      planTasks(db, core);
+    } finally {
+      db.close();
+    }
+  } finally {
+    process.chdir(cwd);
+  }
+
+  return dir;
+}
+
+// Runs the real CLI's `next` in the fixture and returns its stdout.
+// NO_COLOR keeps the output free of ANSI escapes so it can be compared
+// literally.
+export function runNext(dir) {
+  return execFileSync(process.execPath, [CLI, 'next'], {
+    cwd: dir,
+    encoding: 'utf8',
+    env: { ...process.env, NO_COLOR: '1' },
+  });
+}
+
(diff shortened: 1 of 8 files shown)


skyf0xx:
Clean, self-contained, recommended early in merge order. Prompt-level anti-fabrication instruction — the honest admission that verify can't enforce it is the right call.

Needs rebase onto current master.

---

skyf0xx:

---

skyf0xx:
Landed on master at 8174ed0, rebased onto current master. Two conflicts, both in files that render/enumerate the packet's sections — expected, since this PR appends a new one:

- src/db/next.mjs — merged the HONESTY constant and its append-after-VERIFICATION rendering alongside #19's INHERITED DEBT section and #12's incompleteDeps, all now coexisting in formatPacket.
- hedgehog-authored-loop/SKILL.md's packet-section enumeration — combined into STATUS/INTENT/RELEVANT RULES/INHERITED DEBT/WHY NOW/BLOCKED DOWNSTREAM/ALLOWED SCOPE/VERIFICATION/HONESTY.

One test fix, same "stale baseline after an earlier PR landed" pattern seen elsewhere in this queue: repro/packet-sections-preserved.mjs hardcoded the pre-honesty packet as its BASELINE and expected SECTION_ORDER, both frozen at a point before #12's GOAL/OUTCOME labelling and #19's INHERITED DEBT section existed. The test's actual purpose — proving HONESTY is strictly appended after VERIFICATION without disturbing anything else — still holds and still passed; only the literal comparison text was behind. Updated BASELINE and SECTION_ORDER to the current, already-landed format.

Verified before landing: `node scripts/check.mjs` passes, the full 20-script numbered suite passes except the pre-existing, unrelated 05-shipped-cores-unchanged.mjs fixture-staleness failure, and both of this PR's own reproductions pass (packet-honesty-section, and the corrected packet-sections-preserved).

The four observed defensive behaviors motivating this (noisy stub, refuse-to-invent, "unavailable" over fabricated 0, report-don't-widen-scope) are a good, concrete basis for a standing instruction — nice that it's derived from what agents already did right rather than a hypothetical.

---

Issue #5, opened 2026-08-06
hedgehog-core-design: verify command filter tokens aren't cross-checked against the layer's own scope glob


skyf0xx:
When `hedgehog-core-design` authors a layer's `verify` command as a
test-runner filter string (e.g. `pnpm test scale-out` or `pnpm test
ingest calibrate outer-loop`), nothing checks that each filter token
actually corresponds to a real file path inside that same layer's own
`scope` glob list in `core.yaml`. This produced two related but
distinct failure modes in one authored-core build (idea-machine,
data-pipeline shape):

1. **Hard failure**: a filter token that legitimately needed a
   matching test file didn't have that file's path listed in the
   layer's scope, so `hedgehog verify` rejected the layer-eng agent's
   own test file as out-of-scope — even though the verify command
   structurally requires that file to exist to pass. Worked around by
   adding the missing test paths to the layer's scope list by hand.

2. **Silent pass-through**: a filter token (`outer-loop`) that doesn't
   match any filename in that layer's scope at all. The test runner
   (vitest) silently contributes zero tests for a filter that matches
   nothing rather than failing on an empty match set — so the verify
   command *reads* as gating a specific file it structurally never
   forces to run under that filter. No coverage gap materialized this
   time only because the file still ran under other filters in the
   same command, but the command is silently weaker than its own text
   implies.

Both trace to the same root cause: verify-command filter tokens and
scope-glob paths are authored independently, with no check that they
agree.

**Proposed fix**: `hedgehog-core-design` (or a `core.yaml` authoring
lint pass) should cross-check, for every layer, that each filter token
in that layer's `verify` command has at least one corresponding path
in that layer's own `scope` list — flagging both a token with zero
scope matches (silent no-op risk) and a scope-listed test file with no
corresponding filter token coverage, at author time rather than
leaving it to be discovered mid-build by `hedgehog verify` rejections
or worked around live.

Traces to: idea-machine build, friction log entries logged
2026-08-06 08:42:30 (M1-SUBSTRATE-DEDUPE-SCALE-OUT) and 08:53:51
(M1-SUBSTRATE-DEDUPE-OUTER-LOOP), fixed live in commit cf6f643.

skyf0xx:
Fixed in 4b6d086.

Confirmed with `src/db/core.mjs`: `scope` and `verify` are parsed and validated independently (each just needs to be non-empty) — nothing cross-checks a `verify` filter token against the layer's own `scope` globs, matching the report.

[...]

PR #28, opened 2026-08-08
feat(core): require a verify_radius to contain its scope, and flag a verify that never exercises the radius it declares

reinaldosaraiva:
A layer's verify command can be narrower than the `verify_radius` it declares, so a task can break another module's tests and still pass green.

Observed on a real build: a layer declared `verify_radius: ["apps/api/**"]` — correctly, since its command typechecks the whole project — but the test half of its command ran only its own module's specs. Binding a real adapter to a port then invalidated, by construction, a test in *another* module asserting that port was still unbound. That spec lives outside the filter, so it never ran. `tsc` catches a type error but not a false assertion, and the pre-commit hook only typechecks packages with staged files and never runs tests — so the defect crossed both gates. It was caught only because a human hand-wrote the packet with an instruction to also run the neighbouring module's tests.

## How this differs from #5

#5 (fixed in `4b6d086`) checks `verify`'s filter tokens against the layer's **own `scope`**. This axis checks `verify` against **`verify_radius`**, the strictly wider set. A command can satisfy #5 perfectly and still leave everything between scope and radius unexercised. The new rule sits immediately after that one and refers to it rather than restating it.

## A stronger finding underneath, and the hard check it justifies

`verifyRadius()` returns the declared radius **alone**, not unioned with scope:

```js
export function verifyRadius(task) {
  if (task.verify_radius == null) return scope(task);
  return JSON.parse(task.verify_radius);
}

export function conflicts(a, b) {
  if (a.exclusive || b.exclusive) return 'exclusive';
  if (scopesOverlap(scope(a), scope(b))) return 'scope';
  if (scopesOverlap(verifyRadius(a), verifyRadius(b))) return 'verify';
  return null;
}
```

Nothing ever compares one task's *scope* against another's *radius*. So the writes-vs-reads case — A writes a file that B's verify reads — is visible to the scheduler **only because each radius contains its own scope**. The unset default holds that invariant for free; a declared radius narrower than scope silently breaks it, and two tasks that genuinely interfere are handed out together.

`validateCore` never looked at `verify_radius` at all; the parser's only mention is the `null` sentinel.

**Hard check (throws from `validateCore`):** a declared `verify_radius` must contain the layer's own `scope`, and an empty declared radius is rejected separately. Unambiguously wrong, so it fails rather than warns.

## The warning, and the three checks that were rejected

**Warning (`lintCore()`, a new export):** radius strictly wider than scope, **and** every path argument in the verify command sits inside that scope → warn, naming the neighbouring layer whose scope the radius covers. It fires on evidence, never on absence of it.

Against the real authored core it reports four true positives, including:

> layer "api": verify_radius [apps/api/**] is wider than scope, but every path in verify ("src/{module}/http") is inside that scope … The radius covers layer "domain-service"'s scope, which this command never runs.

— naming precisely the neighbour whose spec broke.

Three candidate checks were evaluated and **rejected**, which matters as much as the one kept:

- *"Wider radius and no path-like token outside scope"* — the naive form. It reports on the **absence** of evidence. `pnpm nx test db --testPathPattern={module}` has no path token at all, so it would fire on full-stack-app's own `schema` and `hook` layers — a conclusion the checker has no basis for, since whether a regex filter reaches the whole radius depends on the test corpus, not the core file.
- *"Radius overlaps another layer's scope and the command cannot reach it"* — reachability is the undecidable part, and the observed break was cross-**module** within a layer, not cross-layer. Folded into the warning's message instead of standing as its own check.
- *"Command references a path outside the declared radius"* (the inverse) — a command's path arguments are relative to whatever the runner chdir'd into (`pnpm --filter api exec vitest run src/{module}/http/` is `apps/api`-relative), so a token that fails to resolve under the radius at repo root is evidence of nothing. Real false-alarm risk; dropped.

A lint that cries wolf gets ignored, which would cost more than the checks are worth.

## False-alarm evidence

Both shipped cores load, validate and lint **clean** — and `scripts/check.mjs` now fails on any warning for them, so that stays true. Reproduction cases pin the shapes that must stay silent: a whole-suite command, an ancestor path argument, and no declared radius all produce zero warnings.

## Wiring

- `hedgehog status` — the command every session starts with — prints a `CORE WARNINGS` block, and renders a `validateCore` failure as a one-line `CORE … is invalid:` message instead of dying with a stack trace.
- `scripts/check.mjs` §5 lints both shipped cores as a release gate.
- `globPrefix` is now exported from `conflict.mjs`, so the lint reasons in the scheduler's own glob algebra rather than a drifting copy of it.
- `src/db/verify.mjs` and `src/db/plan.mjs` are untouched.

## The authoring rule

A new Step 5 constraint in `hedgehog-core-design`, answerable yes or no, stating the consequence concretely: a task that breaks a neighbour and still passes. Plus a two-sentence forward pointer at the end of Step 4b, where the radius is first chosen.

## What remains unenforceable, stated plainly

Whether a command *actually reads* its declared radius is undecidable in general. The lint sees only path arguments; a filter expressed as a flag (`--testPathPattern={module}`, `-t <name>`, `--project x`) is invisible to it.

**Worth flagging:** full-stack-app's own `schema` and `hook` layers are structurally this exact shape — a package-wide radius with a flag-narrowed test run — and the lint deliberately abstains on them. Shipped golden-core verify commands were not changed here, because that alters what every full-stack-app project runs at each commit and is a product decision rather than a side effect of this fix. The Step 5 rule is what covers that residue.

## Evidence

`repro/radius-covers-scope.mjs` — before: 2 failures (a narrow radius and an empty radius both "loaded clean", printing the accepted value). `repro/verify-covers-radius.mjs` — before: 1 failure (`no such export (exports: loadCore, parseCoreYaml, validateCore)`). Both exit 1 before, 0 after. `node scripts/check.mjs` passes.

## Conflicts

Textual conflict with **#22** is expected in `hedgehog-core-design/SKILL.md`: the new bullet appends to the same Step 5 constraint list #22 adds to, and the Step 4b pointer sits adjacent to where #22 inserts Step 4c. **#24**'s `requires:` field lands in `core.mjs`'s parser and `validateCore`, which this also edits — adjacent, not overlapping.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -18,7 +18,7 @@ import { fileURLToPath } from 'node:url';
 import { dirname, join, relative, resolve } from 'node:path';
 import { spawn } from 'node:child_process';
 import { dbInit, DB_PATH, openDb } from '../src/db/init.mjs';
-import { loadCore } from '../src/db/core.mjs';
+import { loadCore, lintCore } from '../src/db/core.mjs';
 import { planTasks } from '../src/db/plan.mjs';
 import { addIntent, INTENTS_DIR } from '../src/db/intent.mjs';
 import { nextTask, formatNext, stalledTasks } from '../src/db/next.mjs';
@@ -947,6 +947,34 @@ async function renewCommand(args) {
   console.log(`${green(bold('Renewed.'))} Task ${bold(taskId)}'s lease extended by ${minutes} minute(s).`);
 }
 
+// Heuristic problems with the project's own core definition, rendered for
+// `hedgehog status`. Surfaced here because status is the command every
+// session starts with, and a core-definition smell is a build-wide fact
+// rather than one task's: a layer whose verify command doesn't reach the
+// verify_radius it declared lets a task break a neighbour and still
+// commit green, and nothing later in the loop will say so. Warnings only
+// — the certainties throw from validateCore, on load.
+async function coreWarningLines() {
+  const corePath = await resolveCorePath();
+  if (!corePath) return [];
+  const label = relative(DEST_ROOT, corePath) || corePath;
+  let core;
+  try {
+    core = await loadCore(corePath);
+  } catch (err) {
+    // Every other command that loads the core fails outright on this; say
+    // so here rather than letting `hedgehog status` die with a stack.
+    return ['', `${red('CORE')}  ${bold(label)} is invalid: ${err.message}`];
+  }
+  const warnings = lintCore(core);
+  if (warnings.length === 0) return [];
+  return [
+    '',
+    `${yellow('CORE WARNINGS')}  ${dim(label)}`,
+    ...warnings.map((warning) => `  ${yellow('!')} ${warning}`),
+  ];
+}
+
 async function statusCommand() {
   await ensureDb();
 
@@ -965,6 +993,8 @@ async function statusCommand() {
   }
 
   console.log(formatStatus(result));
+  const warningLines = await coreWarningLines();
+  if (warningLines.length > 0) console.log(warningLines.join('\n'));
 }
 
 // `hedgehog ready` — read-only preview of what a `hedgehog claim` call
--- a/repro/radius-covers-scope.mjs
+++ b/repro/radius-covers-scope.mjs
@@ -21,10 +21,59 @@ import { mkdtemp, writeFile, rm } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join, dirname, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { loadCore } from '../src/db/core.mjs';
+import { loadCore, lintCore } from '../src/db/core.mjs';
 
 const ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '..');
 const failures = [];
+
+// Containment between two globs, in the dialect verify.mjs enforces scope
+// with (git pathspec `:(glob)`): `*`/`?` stay inside one segment, `**`
+// spans any number. A segment-prefix comparison — the scheduler's own
+// globsOverlap algebra — answers "could these two globs ever meet", which
+// is NOT containment: every `reject` row below shares a literal prefix
+// with its scope glob while matching none, some, or the wrong part of it.
+//
+// `accept` rows must load: a wrong hard failure is worse than a missing
+// one, since it breaks every command on a legitimate core.
+const CONTAINMENT = {
+  reject: [
+    // The reviewer's case: a single-segment wildcard cannot reach a
+    // deeper path, however much literal prefix it shares.
+    { scope: ['a/b/**'], radius: ['a/*.ts'], why: 'a/*.ts matches nothing under a/b/' },
+    { scope: ['a/b/**'], radius: ['a/*'], why: 'a/* is one segment deep, a/b/** is deeper' },
+    { scope: ['apps/api/src/**'], radius: ['apps/api/*.ts'], why: 'sibling files, not the tree' },
+    { scope: ['a/b/c.ts'], radius: ['a/*/d.ts'], why: 'literal segment mismatch under the wildcard' },
+    { scope: ['a/**'], radius: ['a/b/**'], why: 'radius is the narrower subtree' },
+    { scope: ['a/b/**'], radius: ['a/b'], why: 'a bare directory path is not its subtree' },
+  ],
+  accept: [
+    { scope: ['a/b/**'], radius: ['a/**'], why: 'trailing ** covers everything below' },
+    { scope: ['a/b/**'], radius: ['a/b/**'], why: 'identical' },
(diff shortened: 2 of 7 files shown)


skyf0xx:
Strongest PR in the queue — merge candidate as-is. Verified at `conflict.mjs:38-52`: `verifyRadius()` returns the declared list alone and is never compared against scope, so a narrow declared radius silently breaks the write/read collision check. Checked against both shipped cores, no false alarms.

Minor: `globCovers`'"'"'s permissive wildcard-tail matching slightly loosens the new hard check too — worth a comment noting it.

---

reinaldosaraiva:
Update: an independent automated review caught a hole in the hard check itself, and it reproduces.

**Wildcard radius bypasses containment.** A core with `scope: ["a/b/**"]` and `verify_radius: ["a/*.ts"]` loaded clean. `globCovers("a/*.ts", "a/b/**")` returned `true`: both globs went through `globPrefix`, which truncates at the first wildcard — `["a"]` and `["a","b"]` — saw a segment-wise prefix, and returned true. **The `*.ts` segment was never examined at all.**

The diagnosis is precise and worth stating, because it was my mistake: prefix comparison answers *"could these two globs ever meet"*, which is all `conflicts()` ever needed, and I reused it for a containment question it cannot answer. Five more shapes had the same hole (`a/*` around `a/b/**`, `a/*/d.ts` around `a/b/c.ts`, `a/b` around `a/b/**`, …). So a core that this PR exists to reject was loading clean.

**Fix:** containment is now decided by walking segments, with an in-segment matcher compiled per pattern, in git's own dialect (`verify.mjs` enforces scope through git pathspecs like `:(glob)packages/db/src/schema/**`, so `*`/`?` within a segment and `**` across any number).

It is **three-valued**, because a wrong hard failure is worse than a missing one:

- **yes** — structurally proven (trailing `**` covers everything below; a literal segment must match; `*` covers any one segment).
- **no** — proven by **witness**: a concrete path the scope glob certainly matches and no radius glob matches. **Only this throws.**
- **unknown** — neither proven. The core loads and `lintCore` emits a `cannot prove …` warning naming the pair.

The witness requirement is what keeps the hard failure honest. The union of radius globs can cover what no single member covers — `["a/b/*", "a/b/*/**"]` together cover `a/b/**` — so "no member contains it" is not on its own grounds to reject a core. Witnesses are also refused (→ unknown) where they cannot be constructed with certainty: interior `**`, character classes.

**Both directions proven.** Now rejected: 6 shapes, all of which loaded clean before, including the reviewer's. Still accepted: 9, including identical globs, `a/**` around `a/b/**`, `{module}` placeholders as literal segments, and the shapes the real cores use. The union case loads with exactly one `cannot prove` warning rather than throwing.

**False-alarm evidence holds:** full-stack-app 0 warnings, landing-page 0 warnings, and the real authored core still reports the same 4 verify-versus-radius warnings as before with **zero** `cannot prove` among them.

Reproductions: 6 failures before, exit 0 after. `scripts/check.mjs` green. The prose rule in `hedgehog-core-design` was corrected to say `validateCore` rejects what *provably* does not cover, and `status` warns where containment cannot be decided.

skyf0xx:

---

skyf0xx:
Landed on master at 523d77b, rebased onto current master. Three conflicts resolved (src/db/core.mjs's validateCore — combined the once/{module} checks with the new verify_radius-contains-scope check in the same per-layer loop; bin/cli.mjs — combined the lintCore import and appended CORE WARNINGS reporting alongside #20's commit-gate reporting in statusCommand; hedgehog-core-design/SKILL.md — sequenced this PR's Step 5 paragraph before #22's Step 4c section, both prose-only additions).

Your own self-correction in the second commit (deciding radius containment via the actual git-pathspec dialect instead of a prefix comparison that couldn't distinguish "shares a prefix" from "actually covers") held up under review — good catch on the false-green that prefix comparison would have let through.

One real finding from running this PR's own release-gate lint against current master, not a merge-mechanics issue: both shipped full-stack-app layers (schema, hook) failed the new hard check, expecting 0 warnings and getting 2. Traced it to a300678 (already on master, unrelated to this PR or any other in this queue) — it anchored these layers' verify commands to their own module's directory to fix a Jest-flag-on-Vitest bug, but under this workspace's Nx target inference, `test` and `typecheck` are distinct targets, so the anchored command no longer typechecks the whole package the way the original (pre-a300678) verify_radius rationale assumed. Since a300678's own commit message explicitly deferred wiring in this exact lint, this reads as the gap that commit expected a follow-up to close. Added a separate commit narrowing verify_radius on both layers back to match scope, with an updated core.yaml header comment explaining why — kept out of your two commits since it's fixing pre-existing drift the lint surfaced, not part of this PR's contribution.

Verified before landing: `node scripts/check.mjs` passes clean (0 warnings on both shipped cores), the full 20-script numbered suite passes except the pre-existing, unrelated 05-shipped-cores-unchanged.mjs fixture-staleness failure (a stale JSON snapshot predating a300678's command-format change, now further out of date after the verify_radius fix — worth a fixture regen separately), and both of this PR's own reproductions pass.

---

PR #29, opened 2026-08-08
feat(plan): add layer cardinality so cross-cutting work compiles once, not once per module

reinaldosaraiva:
A module-axis core cannot express a layer that should run **once**. `compileIntentTasks` does `core.layers.map(...)` for every intent, and no layer field controls cardinality — `exclusive: true` only affects scheduler concurrency, not how many tasks are compiled.

Measured on a real 6-module build: cross-cutting work — cluster provisioning and deployment — compiled **six times each**. Ten of those twelve executions were pure replication of `terraform apply` and `kubectl apply` against a live cluster, each `exclusive` and therefore stopping the whole build. Nothing corrupted (terraform has state, `kubectl apply` is idempotent per object), but it cost minutes of redundant serialized work and made the commit log attribute infrastructure to an arbitrary module.

## A contradiction already in the payload

`hedgehog-core-design`'s Step 4 tells the authoring agent to add cross-cutting work as

> a layer whose `scope` is a fixed path with no `{module}` placeholder

while `validateCore` rejects precisely that:

```js
if (isModuleAxis) {
  for (const layer of core.layers) {
    if (layer.exclusive) continue;
    if (!layer.scope.join('').includes('{module}')) throw new Error(...)
```

So today the shipped guidance produces either a load error, or — if the author also sets `exclusive: true` to get past the check — the six-fold replication being complained about. There is no third outcome.

## Design decisions

**Task id.** A once-layer's task id is the bare uppercased layer id — `CLUSTER`, `DEPLOY`. The absent intent prefix *is* the signal. A collision with a real `<INTENT>-<LAYER>` id is caught at compile with a named error rather than surfacing as a mid-transaction primary-key failure.

**`intent_id`, and why not a schema change.** `tasks.intent_id` is `NOT NULL REFERENCES intents(id)`, so a once-task needs an owner. Making the column nullable was rejected: `applySchema` is `CREATE TABLE IF NOT EXISTS` and there is no migration mechanism, so existing databases would stay on the old schema — that is a maintainer's call, not a contribution.

Attributing it to a real intent was also rejected, on two grounds. It links infrastructure to one module's requirements and holds that intent open via `completeIntentIfDone`; and — decisively — **it is not reproducible.** `planTasks` compiles pending intents in `priority, id` order, so a once-task attributed to "zeta" live becomes attributed to "alpha" after a `db rebuild` that replays both. Since the whole promise is that the graph is derived from committed sources, that is disqualifying.

Instead the compiler synthesises a `_core` intent derived entirely from `core.yaml` (goal and outcome from the core id, status `active` so it is never picked up as pending). `module` is `_core` too, and `intent add --id _core` is now refused. Reproduction 06 pins this with a deliberately inverted add order.

**Dependency edges, both directions.** per-module → once compiles one edge onto the single task. once → per-module compiles one edge per intent, accumulating into "waits on all of them" and growing correctly when a later intent is added. once → once is compiled in the once-pass.

Also found and handled: with a once-layer present, the cross-intent edge loop would compile a once-task edge **to itself**, which the `dependencies` `CHECK (task_id <> depends_on_task_id)` rejects outright.

Documented limitation: adding a once-layer to a core whose intents are already compiled leaves those skipped intents contributing no edges. The fix is to delete the DB and `db rebuild`; it is stated in a code comment rather than left to be discovered.

**`{module}` in a once-layer** is a hard validation error naming the layer, the placeholder and the reason. Silent mis-substitution would produce a glob matching a literal `{module}` directory and — worse — a commit message that `db rebuild` matches git history against.

**Incremental compile.** The already-compiled check now keys on the first *per-intent* layer. Keying on `core.layers[0]` with a once-layer at the head would give every intent the same probe id, so every intent after the first would look compiled and silently compile nothing. `validateCore` rejects an all-`once` core, which guarantees a per-intent layer exists.

## `applies_to` — deliberately not implemented

It is a different axis (selection, not cardinality) and it breaks an invariant `once` does not: naming intent ids inside `core.yaml` makes a locked, intent-agnostic core definition intent-*aware*, so every re-entry pass adding a module would require editing a locked file — turning ordinary new scope into a Correction Protocol event.

It also raises a chain-splicing question with no obviously right answer: if B `depends_on` A and A does not apply to intent I, does I's B depend on nothing, or on A's predecessor? That changes the meaning of `depends_on` for every existing core. `once` covers the measured cost; `applies_to` deserves its own proposal.

## The scheduler half

`conflicts()` returns `'exclusive'` if *either* side is exclusive, and `claimTasks` only accepts a candidate whose conflict set is empty — so an exclusive candidate is claimable only when it is the first candidate `ORDER BY t.priority, t.id` hands over. Since every completing task opens another in the same module, a lower-sorting non-exclusive candidate is nearly always available.

**One correction to how this was reported to me:** it is *indefinite deferral*, not unbounded starvation. In a finite graph the exclusive task is eventually claimed once every lower-sorting non-exclusive candidate is exhausted. The practical severity stands — as a module's first layer it stalls that module for the rest of the build — but "starved forever" overstates it.

Fix: `ORDER BY t.priority, t.exclusive DESC, t.id`, mirrored across `claim.mjs`, `next.mjs` and `status.mjs` so `ready`/`next`/`status` keep agreeing with what `claim` would actually take. `conflicts()` is untouched, and no aging priority was added — that would need a "claimable since" measure the schema does not carry, and would mutate graph state.

The throughput argument: an exclusive task runs alone whenever it runs, so moving it earlier cannot cost aggregate time — it moves the same serialized step forward and unblocks its dependents sooner. Reproduction 07's third assertion pins that non-exclusive fan-out is otherwise unchanged.

## Evidence

`node --experimental-sqlite repro/run-all.mjs`:

| | before | after |
|---|---|---|
| 01 `once` compiles one task | FAIL — got `BILLING-CLUSTER`/`ORDERS-CLUSTER`/`USERS-CLUSTER` | PASS |
| 02 edges in both directions | FAIL — `USERS-SCHEMA → USERS-CLUSTER`; `DEPLOY` deps `[]` | PASS |
| 03 a fourth intent adds no duplicate | FAIL — 16 tasks including 4 CLUSTER + 4 DEPLOY | PASS |
| 04 `{module}` in a once-layer is rejected | FAIL — exits 0, "1 intent(s) compiled" | PASS |
| 05 shipped cores compile unchanged (**control**) | PASS | PASS |
| 06 survives `db rebuild` | FAIL — `ALPHA-CLUSTER`/`ZETA-CLUSTER` | PASS |
| 07 exclusive task not deferred | FAIL — claimed `ALPHA-SCHEMA`, not `ZULU-PROVISION` | PASS |

6 of 7 fail before; all 7 pass after. **05 is a control and passes on both sides by design** — "compiles byte-identically to before" cannot be a failing reproduction. Its baseline was generated by running the same script inside a clean clone of HEAD, and compares the full `intents`/`tasks`/`dependencies`/`task_requirements` dump for both shipped cores with two intents, a dependency edge and a requirement.

End-to-end smoke against a real git repo: a single `CLUSTER` task → `claim` → `verify` → one commit `chore(infra): cluster` attributed to infrastructure rather than a module → both modules' schema tasks unlocked → `db rebuild` reconciles it back to `complete` from the commit subject. `hedgehog next` on a once-task prints `✓ Cross-cutting layer — one task for the whole build` instead of naming a fake module.

`node scripts/check.mjs` passes. `src/db/verify.mjs` untouched.

## Conflicts

- **`src/db/plan.mjs` vs #17** — near-certain, and **substantive rather than textual**. #17 extracts `layerTaskFields()`, which is exactly the function `compileOnceTasks` needs a non-`{module}`-substituting variant of; and `plan --recompile` will have to decide whether recompiling drops and reinstates the single once-task. Flagging explicitly rather than leaving it to the merge.
- **`src/db/core.mjs` vs #24** — textual. #24's `requires:` and this `once:` both add a field to the `parseCoreYaml` literal and a block to `validateCore`; independent but adjacent.

## One note on scope

`hedgehog-core-design`'s SKILL.md is edited here to document `once: true` and to fix the Step 4 guidance quoted at the top. Without it the field is unreachable by the only agent that authors core definitions — but it is payload rather than engine, so say the word if you would rather have those split.


skyf0xx:
Clean, resolves a real contradiction (core-design tells authors to write fixed-scope layers; validateCore rejects them). Already based on d43df75, thanks for rebasing.

**Same #17 collision noted there**: your `_core` synthetic module won'"'"'t match #17'"'"'s `layerTaskFields` lookup — once-tasks would drift forever if #17 lands first unadjusted.

Also collides directly with #24 in `core.mjs`'"'"'s `parseCoreYaml` layer literal — second one in is a hand-merge.

---

reinaldosaraiva:
Update: an independent automated review of this branch (Greptile, on a fork where it could run) raised a P1 I had missed, and it reproduces — worse than reported.

**Completed once-task gains prerequisites.** When a new intent is planned after a tail once-layer such as `deploy` has completed, the new dependency edge is created but the once-task stays `complete`. SQLite has no constraint against adding a prerequisite to a complete task, and the readiness subquery only gates tasks that are not already complete — so the edge exists and is never consulted.

Reproduced through the full Re-entry flow with real commits:

```
hedgehog status  →  TASKS  8 / complete  8 / READY (none)
hedgehog next    →  No ready task. Nothing is planned with all dependencies complete.
```

The graph declares the build finished with the new module never deployed — through `planner`'s Re-entry pass, which is a documented first-class flow, not an edge case.

**Fixed rather than documented**, because re-running is what a cross-cutting layer is *for*: these commands are idempotent by construction, so re-running the deploy after a module lands is the semantically correct action.

`hedgehog plan` now sets a completed once-task back to `planned` when a newly compiled intent hangs a prerequisite on it, printing `reopened  DEPLOY (once — new scope landed under it; it must run again)`, cascading along once→once edges. The earlier commit stays in history untouched — this moves status, not the record.

**The `db rebuild` interaction turned out to be the harder half**, and the reopen was worthless without it. A rebuild re-closed the task from the old commit, because completion is inferred from commit-subject *membership* and a once-layer's subject carries no `{module}` — it is a constant, so one historical `chore(infra): deploy` makes the layer look done forever. Membership cannot express "ran after". `loadCommitSubjects` now records each subject's newest position (`git log --topo-order`), and a once-task is complete only if its own commit postdates every prerequisite's. Still derived purely from git and the committed intents — no new state, and not a ratchet: the reproduction commits the deploy again and asserts the graph goes fully green across a rebuild.

Scoping, since it is the risky part: the postdating rule applies **only** to once-tasks (a task that touches no files leaves no commit at all, so a general rule would cascade that pre-existing gap through per-module chains); per-module tasks downstream are **not** reopened; and replanning while the once-task is leased is refused outright rather than guessed at.

`src/db/rebuild.mjs` now joins the conflict list alongside the `plan.mjs` overlap with #17.

Suite: 9/9 reproductions pass, `scripts/check.mjs` green, shipped cores still compile identically.

PR #48, opened 2026-08-09
ci: run the reproductions, and stop the runner from skipping 31 of them

reinaldosaraiva:
The reproductions in `repro/` are the evidence behind every bug fix that
landed in this repo over the last few days. None of them run in CI.

`check.yml` and `publish.yml` each run `npm run check` and nothing else, and
that script validates the payload — agent and skill frontmatter,
cross-references, the two shipped cores, the tarball contents, the CLI
entrypoint. It never touches `repro/`. So the suite that proves the fixes work
is a gate that exists and does not run.

Three defects here, and they hold each other up.

### 1. The runner discovers 20 of 51

`run-all.mjs` selects on `/^\d+-.*\.mjs$/`. Every reproduction whose author
gave it a descriptive name instead of a number is skipped in silence — 31 of
them, including the yaml-scalar-escape, no-op-layer-commit, `boundary`,
packet-goal, packet-honesty, verify-radius, drift and add-intent-write-failure
suites. You noted this yourself when landing #18 (*"your four boundary-*.mjs
scripts aren't digit-prefixed so they don't match the glob"*).

Demonstrated rather than asserted. Commenting out `formatPacket`'s
`lines.push(...HONESTY)` — deleting the packet section added in `8174ed0` —
leaves the old runner reporting **20 of 20 passed, green**. Under this branch
the same tree fails `packet-honesty-section` and `packet-sections-preserved`,
and exits 1.

The fix inverts the rule to a denylist of helpers. An allowlist fails silently
when a new reproduction doesn't match the pattern; a denylist fails loudly,
because a new helper gets executed as a reproduction and errors. 51
reproductions, 10 helpers, nothing in neither set.

### 2. `05-shipped-cores-unchanged` has been red since `a300678`

It was waved through as "pre-existing and unrelated" on essentially every pull
request in this queue, including mine. It is neither a flake nor a regression:
the baseline froze the full-stack-app core's *compiled verify commands*, and
two later commits legitimately changed them — `a300678` replaced six Jest
`--testPathPattern=` flags with Vitest positional filters, and `523d77b`
removed schema/hook's `verify_radius` once those anchored commands stopped
reading the whole package.

This is the third defect's cause. A command that is permanently red teaches its
reader to ignore the exit code, and from there nobody notices that it also
doesn't run in CI and discovers half the files.

Regenerated the way the script's own header requires — from a checkout that
predates the cardinality feature (`e8d1571^`), with today's `core.yaml` copied
in, so the baseline still records what the *pre-cardinality* compiler produced
and only the core.yaml-derived fields move. Diff is 16 lines, all in
full-stack-app's verify commands and `verify_radius`; `landing-page` is
byte-identical, and task ids, dependency rows and requirement links are
unchanged. Regenerating from master instead would have made the assertion
compare current output against current output and prove nothing.

### 3. Nothing runs the suite

`npm run repro`, wired into `check.yml`. **18 seconds for all 52** on a warm
cache.

## Verified

On a clean clone of `master` at `dc730fc`:

| | |
|---|---|
| `npm run check` | passes, unchanged |
| `npm run repro` | **51/51**, exit 0, 18s |
| with an injected regression | exits 1, names the failures |
| old runner, same regression | 20/20 green |

## One call for you

Two reproductions (`first-commit-rejected`, `commit-gate-fails-open`)
install the pinned lefthook from the registry into a
throwaway directory, and declare that in their own headers. That install
belongs to the reproduction, not to the workflow: `--ignore-scripts` still
governs this repository's own dependencies, and the job keeps its read-only
token and its `pull_request` trigger.

But it is a change in what a fork's pull request causes to happen on your
runner, and that was a deliberate design in `check.yml`'s header comment. If
you'd rather those three didn't run on fork PRs, they're separable into their
own step or gated on `github.event_name != 'pull_request'` — say the word and
I'll split them.

Unrelated and not fixed here: root `node_modules/` isn't in `.gitignore` (only
the golden cores' are), so a local `npm install` leaves it untracked.

---

Independent review of this branch is in flight in a parallel fork with Greptile
installed, for the same reason as last time — contributor branches aren't run
here, so someone who didn't write it should read it. I'll report anything it
finds.


reinaldosaraiva:
Fourth commit added: `fix(repro): take the shell out of the boundary stderr capture`.

CodeQL on a mirror flagged `repro/lib/fixture.mjs:79` — a shell command built from environment values — and it is right. `runCliCapturingBoth` built `<node> <cli> <args> 2><errPath>` as a shell string and escaped each part with `JSON.stringify`.

That is not a shell escape. Inside double quotes `sh` still expands `$(...)` and backticks. Reproduced against the real `/bin/sh`: a capture path containing `$(touch SUBSTITUTED)` runs it.

**This is the third instance of the pattern here.** 7450c84 fixed it in the CLI; you blocked #14 before merge for reintroducing it via `JSON.stringify(commitMessage)`; this one arrived in test code with my own #18 and nobody — including me — looked. Reachable only when TMPDIR or the checkout directory carries shell metacharacters, so the severity is low, but it is code your runner now executes on every pull request *because of this PR*, which is why it belongs here rather than in a follow-up.

Replaced with `execFileSync` and stderr on a file descriptor: same two-stream separation, no shell.

`repro/boundary-capture-no-shell.mjs` pins the premise against the real shell, the replacement, and the helper itself driven through a poisoned TMPDIR.

One note on that third section, because the obvious version was vacuous and I nearly shipped it. Passing TMPDIR as the helper's `env` option makes every assertion pass against the **unfixed** helper — `os.tmpdir()` reads the calling process's environment, not the child's. It has to be set on the repro process. Verified in both directions: exit 1 against the unfixed helper, exit 0 against the fixed one.

52/52 pass, `npm run check` unchanged.

reinaldosaraiva:
Correction, from the independent review on the mirror: **two** reproductions reach the registry, not three.

`commit-failure-recovery` writes its own rejecting pre-commit hook (`#!/bin/sh ... exit 1`) and never invokes npm. The word "lefthook" appears in it only inside the message that hook echoes.

I had grepped the reproductions for `lefthook` and reported the hits as behaviour — a name matched, so I wrote down a claim about what the code does. That is the same error this repo's issue #9 is about, committed while describing the fix for it.

Fixed in `ci: correct which reproductions reach the registry`; the PR body above is updated. `first-commit-rejected` and `commit-gate-fails-open` are the two, and every other reproduction is hermetic.

PR #52, opened 2026-08-09
fix(ci): read the PR diff via the API, not a fork checkout under pull_request_target

### LINKED ISSUE #51: [bug] pr-risk.yml runs on pull_request_target and checks out the fork's merge ref, so every fork PR fails execution-risk
## What happens

Every pull request from a fork fails the `execution-risk` check, regardless of its contents. The check never runs its actual logic — `actions/checkout` refuses before it starts.

Reproduced on #48 (a fork PR), whose `execution-risk` run ends with:

```
allow-unsafe-pr-checkout: false
##[error]Refusing to check out fork pull request code from a 'pull_request_target' workflow.
This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache
scope, and runner access. Fetching and executing a fork's code in that trusted context
commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at
https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true'
on the actions/checkout step.
```

## Root cause

`.github/workflows/pr-risk.yml` runs `on: pull_request_target` (so its jobs get `pull-requests: write` to apply risk labels), and its `execution-risk` job does:

```yaml
    steps:
      - uses: actions/checkout@v7
        with:
          persist-credentials: false
          fetch-depth: 0
          ref: "refs/pull/${{ github.event.pull_request.number }}/merge"
```

Checking out a fork's PR merge ref inside a `pull_request_target` job is the classic "pwn request" pattern, and `actions/checkout@v7` refuses it by default (`allow-unsafe-pr-checkout: false`). `actions/checkout` was bumped 5 → 7 in #33, which is when this default started biting. So the job aborts at the checkout step for every external contributor.

## Why it's worth fixing

Two reasons, and the second is the one that matters:

1. **The check is dead for exactly the PRs it exists to label.** Internal pushes are fine; every fork PR — the ones a risk-labelling gate is most for — gets a hard red before the scan runs.

2. **A red `execution-risk` is indistinguishable from a bad PR.** A first-time external contributor sees a failing check named "execution-risk" on their PR and, without access to the run logs (or the knowledge to read them), reasonably concludes the problem is *their* code. The check meant to help triage contributions instead discourages them, silently.

There's also a documentation angle: `.github/workflows/check.yml`'s own header comment states the rule this workflow breaks —

> Keep it on `pull_request` — `pull_request_target` would hand the fork's code a token with write scope.

The repo wrote down the constraint, with the right rationale, and the next workflow added did the opposite. (It fails *safe* only because `checkout@v7` refuses — the configuration itself is the anti-pattern the comment warns against.)

## Suggested fix

The `execution-risk` job needs the PR diff and `pull-requests: write` to label — it does **not** need to check out, let alone execute, fork code. Get the diff as data instead of as a working tree:

- Drop the `actions/checkout` of the fork merge ref entirely.
- Read the changed lines via the API — `gh pr diff ${{ github.event.pull_request.number }}` or `gh api repos/${{ github.repository }}/pulls/<n>/files` — and run the same `child_process|execFile|execSync|spawn(|exec(|eval(|new Function(|Docker` grep against the added lines.

That keeps the `pull-requests: write` scope for labelling while never placing fork-controlled code in the trusted context — which is what `pull_request_target` is safe for and `pull_request` is not.

(If a checkout is genuinely wanted for some future check, check out the **base** repo's ref, not the fork's merge ref, and diff against the head SHA the event already carries.)

## Reproduction

Open any pull request from a fork against `master` (e.g. #48) and observe `execution-risk` fail at the `actions/checkout` step with the message above. No PR content triggers or avoids it.

reinaldosaraiva:
Fixes #51.

## The bug

`execution-risk` in `.github/workflows/pr-risk.yml` runs under `pull_request_target` and checks out the fork's PR merge ref:

```yaml
- uses: actions/checkout@v7
  with:
    ref: "refs/pull/${{ github.event.pull_request.number }}/merge"
```

`actions/checkout@v7` refuses that by default (the pwn-request guard), so since the checkout bump in #33 the job fails at the checkout step on **every** fork PR — before its scan runs. The red is indistinguishable from a bad change, and it's the exact pattern `check.yml`'s own header warns against.

## The fix

The job needs the diff and `pull-requests: write` to apply `risk:execution`. It does **not** need to check out — let alone execute — fork code. Read the diff as data instead:

```yaml
ADDED=$(gh api --paginate "repos/$REPO/pulls/$PR/files" \
  --jq '.[] | select((.filename | startswith("vendor-skills/")) | not) | .patch // ""' \
  | grep -E '^\+' | grep -vE '^\+\+\+' || true)
```

Same token regex, same `vendor-skills` exclusion (now a `jq` filter instead of the `:!vendor-skills` pathspec), no working tree. No fork-controlled code enters the trusted context, and the `pull_request_target` token keeps its write scope for labelling.

## Verification — and why it can't be on this PR

**`pull_request_target` always evaluates the workflow from the base branch**, so this PR's own `execution-risk` still runs master's broken version. The fix can only be observed once it's on `master`. That's a property of the event, not a gap in the change.

So I verified the scan logic directly, running the exact `gh api pulls/<n>/files` pipeline locally against real PRs in this repo:

| PR | Contents | Result |
|----|----------|--------|
| #48 | adds `execFileSync`, `child_process` | `match=true` ✅ |
| #43 | docs (contributing) | `match=false` ✅ |
| #46 | docs (cleanup) | `match=false` ✅ |

The added-line extraction (`^+`, minus `+++` headers) is identical to the old `git diff` version — only the diff's *source* changed from a checkout to the API, so the verdict is unchanged for any given diff.

`pr-risk.yml` validated with `js-yaml`; no `checkout` step remains in the file. No other job in it checks out fork code (`path-based` uses `actions/labeler`, `size` uses `actions/github-script`, both API-only).

---

Independent review in flight on a parallel fork with Greptile, per the usual flow here — contributor branches aren't run upstream, so someone who didn't write it reads it first.

### DIFF
--- a/.github/workflows/pr-risk.yml
+++ b/.github/workflows/pr-risk.yml
@@ -40,16 +40,27 @@ jobs:
           REPO: ${{ github.repository }}
           PR: ${{ github.event.pull_request.number }}
         run: |
-          # Per-file patches, vendor-skills excluded to match the previous
-          # ':!vendor-skills' pathspec. Added lines start with a single
-          # '+'; the '+++' filter drops the file headers. Context width is
-          # irrelevant to that test, so no --unified=0 is needed. `.patch`
-          # is absent for binary/oversized files, hence `// ""`.
-          ADDED=$(gh api --paginate "repos/$REPO/pulls/$PR/files" \
-            --jq '.[] | select((.filename | startswith("vendor-skills/")) | not) | .patch // ""' \
+          FILES=$(gh api --paginate --slurp "repos/$REPO/pulls/$PR/files")
+
+          # Added lines from files whose patch is present, vendor-skills
+          # excluded to match the previous ':!vendor-skills' pathspec.
+          # Added lines start with a single '+'; the '+++' filter drops the
+          # file headers, so context width is irrelevant and no
+          # --unified=0 is needed. (--slurp wraps the paged responses as an
+          # array of arrays, hence '.[][]'.)
+          ADDED=$(echo "$FILES" | jq -r '.[][] | select((.filename | startswith("vendor-skills/")) | not) | .patch // ""' \
             | grep -E '^\+' | grep -vE '^\+\+\+' || true)
-          if echo "$ADDED" | grep -qE 'child_process|execFile|execSync|spawn\(|\bexec\(|\beval\(|new Function\(|Docker(file)?\b'; then
+
+          # GitHub omits `patch` for oversized (or binary) diffs. A text
+          # file with real additions but no patch is one we cannot inspect
+          # — so fail safe and let it count as a match rather than skip it
+          # silently, the way the old whole-tree `git diff` never could.
+          # additions>0 keeps pure deletions/renames from tripping this.
+          BLIND=$(echo "$FILES" | jq -r '.[][] | select((.filename | startswith("vendor-skills/")) | not) | select(.additions > 0 and .patch == null) | .filename')
+
+          if [ -n "$BLIND" ] || echo "$ADDED" | grep -qE 'child_process|execFile|execSync|spawn\(|\bexec\(|\beval\(|new Function\(|Docker(file)?\b'; then
             echo "match=true" >> "$GITHUB_OUTPUT"
+            [ -n "$BLIND" ] && echo "::notice::risk:execution set conservatively — no diff available to scan for: $(echo "$BLIND" | tr '\n' ' ')"
           else
             echo "match=false" >> "$GITHUB_OUTPUT"
           fi
--- a/.github/workflows/pr-risk.yml
+++ b/.github/workflows/pr-risk.yml
@@ -40,7 +40,14 @@ jobs:
           REPO: ${{ github.repository }}
           PR: ${{ github.event.pull_request.number }}
         run: |
+          # The whole diff, as data. The API can hand us less than the full
+          # picture two ways — an omitted patch, or a truncated file list —
+          # and both are handled by failing safe below, so a blind spot
+          # becomes an over-label, never a silent miss. This is the parity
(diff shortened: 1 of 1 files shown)


reinaldosaraiva:
Second commit: `fix(ci): fail safe when github omits a file's patch, don't skip it`.

Independent review caught that `.patch // ""` silently drops a file whose diff GitHub omits for size — so a `risk:execution` token in a large added file would go unlabelled. The old whole-tree `git diff` had no such blind spot, so it was a real coverage regression in my first commit.

Now a non-excluded file with `additions > 0` and no patch forces `match=true` (with a `::notice::` naming the files) instead of being skipped. `additions > 0` keeps pure deletions/renames from tripping it. A risk labeller should over-label when blind, never under-label — this restores parity with the old behaviour. Verified locally against #48 (true), #43/#46 (false), and a synthetic oversized-file case.

reinaldosaraiva:
Third commit: `fix(ci): fail safe when the files endpoint truncates a huge pr`.

Independent review's second pass noted that `pulls/<n>/files` caps at ~3000 files, so a PR larger than that returns only a prefix and a token in the omitted tail would go unlabelled — again a blind spot the old whole-tree `git diff` didn't have.

Closed the class, not the instance: compare the PR's `changed_files` count to how many files the endpoint returned; if fewer, fail safe to `match=true` with a `::notice::`. That covers the 3000-file cap and any pagination truncation in one guard. Between this and the omitted-patch fail-safe, 'can I see the whole diff?' is now a single guarded question. Verified locally; verdicts on #48/#43/#46 unchanged.

skyf0xx:
Looks good - thanks for spottin gthis

PR #132, opened 2026-08-14
fix: work the inbound bug queue — eight issues across runtime, build graph, and generators

### LINKED ISSUE #121: Newly-installed agents are not registered in the session that ran `init`
Installed Hedgehog with `npx @skyf0xx/hedgehog init --ts-full-stack-app`, then followed root `CLAUDE.md`'s fresh-install instructions in the same session.

## What happened

`CLAUDE.md` ("First message in a fresh install") and `planner.md` (step 9) both instruct the session to hand off to the `bootstrap` agent once planning intake commits. That dispatch fails:

```
Agent type 'bootstrap' not found. Available agents: claude, claude-code-guide,
Explore, general-purpose, Plan, planner, qa-tester, reviewer, searcher,
statusline-setup, ui-builder
```

The host enumerates its agent list at session start. `init` writes `.claude/agents/*.md` mid-session, so none of the Hedgehog agents (`bootstrap`, `backend-eng`, `front-end-eng`, `ux-planner`, `reviewer`, `tweaker`, `layer-eng`, …) are dispatchable until the session restarts.

## Why this is worse than a missing agent

The names that *did* resolve were my own global `~/.claude/agents/` definitions — an unrelated `planner` and `reviewer` with different instructions. A `planner` dispatch would have silently run the wrong discipline rather than erroring, which is the failure mode with no symptom.

Note this is the exact case `CLAUDE.md` already anticipates in spirit:

> This same rule holds at every later handoff to a newly-installed agent or skill in this session: read its file, don't assume its content.

Reading the file works. *Dispatching* to it does not.

## Workaround

Read `.claude/agents/bootstrap.md` and follow it inline. This worked fine — the file is self-contained — but it means the whole build ran in one context rather than the isolated per-agent contexts the design intends, which undercuts the "conversation is disposable" property.

## Suggested fix

Any of:

- Have `init` print, as the final "Next steps" line, that the agent set only becomes dispatchable in a **new** session — and that the first build session should be started fresh rather than continuing the install session.
- Have `CLAUDE.md`'s fresh-install section say the `bootstrap` handoff may need to be followed inline if the dispatch fails, naming the symptom so it is not mistaken for a broken install.
- Detect it: if `.claude/agents/bootstrap.md` exists but the dispatch errors, that is a session-staleness signal worth naming explicitly.

Environment: Claude Code (VSCode extension), macOS (Darwin 25.3.0), Hedgehog full-stack-app core.

### LINKED ISSUE #122: Intent id silently decides every scope glob, but the plural-module convention is only documented at the generator
On a first run I wrote a single intent with `id: "task"` (singular — matching how I had phrased the goal). `hedgehog plan` compiled all 8 layers with `{module}` = `task`:

```
ALLOWED SCOPE
  packages/db/src/schema/task/**
  packages/db/src/schema/index.ts

VERIFICATION
  pnpm nx test db -- src/schema/task/
```

Then the layer packet told me to scaffold from:

```
nx g ./tools/generators:schema --module=task --fields=...
```

`hedgehog-loop`'s "Scaffolding a layer" section says `--module` is **"the domain module's plural kebab-case name (`tasks`, `order-items`)"**. So the correct generator invocation is `--module=tasks`, which writes `packages/db/src/schema/tasks/` — outside the compiled scope `packages/db/src/schema/task/**`.

## The problem

The intent id is the single input that decides every layer's scope glob and verify command, and it is chosen at planning intake — several steps before anything states that the module axis is expected to be plural. By the time the convention appears (in the loop skill, at scaffold time), the graph is already compiled against the singular.

Nothing catches it: `intent add` accepts it, `plan` compiles it, `claim` hands out a packet whose `scaffold from:` line is itself the wrong command.

## Recovery cost

Since `.hedgehog/hedgehog.db` is derived and gitignored, the fix is at the committed source:

```bash
git mv .hedgehog/intents/task.json .hedgehog/intents/tasks.json
# edit "id" and every "TASK-*" requirement id
rm -f .hedgehog/hedgehog.db*     # see linked issue — rebuild alone is not enough
hedgehog db rebuild
```

Cheap only because I caught it on the very first generator call, with nothing built. Caught three layers later it would be a Correction Protocol case across every compiled task.

## Suggested fix

Any of:

- Have `hedgehog intent add` warn (or refuse) when an id on a module-axis core is not plural — it already knows the core has a module axis.
- Have `planner.md` state the convention where the id is actually chosen, not only where it is consumed. `planner.md`'s "one table = one module literally" guidance is about *decomposition*, not naming.
- Derive the `scaffold from:` line from the same string as the scope globs so they cannot disagree — currently the packet prints a generator command that contradicts its own ALLOWED SCOPE.

The third seems strongest: the packet telling you to run a command that violates its own scope is the actual defect.

### LINKED ISSUE #123: `hedgehog db rebuild` leaves stale task rows behind instead of replacing the graph
While correcting an intent id (see linked issue), I renamed `.hedgehog/intents/task.json` → `tasks.json`, updated its `id` and requirement ids, and ran:

```bash
hedgehog db rebuild
```

Output:

```
rebuilt  1 intent(s) replayed, 0 task(s) marked complete
```

But `hedgehog status` then showed **16 tasks**, not 8:

```
TASKS  16

  planned      15
  ready        1

READY
  TASK-SCHEMA    schema   schema for task
  TASKS-SCHEMA   schema   schema for tasks   (held back — conflicts with TASK-SCHEMA (scope))
```

The rebuild replayed the corrected intent additively on top of the existing rows rather than rebuilding the graph from the committed sources. The old `TASK-*` tasks — whose intent file no longer exists — survived, and the scheduler then correctly detected that the stale `TASK-SCHEMA` and the new `TASKS-SCHEMA` have overlapping scope, holding the real task back behind a ghost.

## Workaround

Delete the derived DB first, then rebuild:

```bash
rm -f .hedgehog/hedgehog.db .hedgehog/hedgehog.db-shm .hedgehog/hedgehog.db-wal
hedgehog db rebuild
```

→ `TASKS 8`, clean, `TASKS-SCHEMA` ready.

## Why this matters beyond my case

`CLAUDE.md` presents rebuild as the thing that makes the DB safely disposable:

> `.hedgehog/hedgehog.db` is gitignored: a derived index, rebuildable at any time via `hedgehog db rebuild`, which replays those committed sources against git history.

"Rebuildable at any time" implies the result is a pure function of the committed sources. Here it is a function of the sources *plus whatever was already in the DB*, so a rebuild over an existing DB can produce a graph that no set of committed intents describes. The auto-rebuild on fresh clone is fine (no prior DB); the manual path is where this bites.

## Suggested fix

Make `db rebuild` truncate the derived tables before replaying — it is by definition reconstructing everything it needs from `.hedgehog/intents/` plus git history. Failing that, a `--clean` flag, or at minimum have it drop task rows whose intent no longer exists in `.hedgehog/intents/`.

### LINKED ISSUE #124: First-arrival override is documented as pre-emptive but is only discoverable after a failed verify, costing a 5-command recovery
`hedgehog-loop`'s "First arrival in a package" section says:

> Widen that one task, **before building it**:
> ```bash
> hedgehog override add TASKS-CONTRACT --scope 'packages/contracts/*' ...
> ```

The difficulty is that nothing tells you the task is a first arrival *until you have already built it*. The packet for `TASKS-CONTRACT` prints its ALLOWED SCOPE (`packages/contracts/src/{module}/**`) and a `scaffold from:` line — but the generator, correctly, also creates the package shell (`package.json`, `tsconfig*.json`, `vitest.config.mts`, `src/index.ts`, `src/timestamp.ts`), all outside that scope. You learn this from `hedgehog verify` rejecting 9 paths.

By then the task is `blocked`, and "widen it before building" is no longer available. The recovery is five commands, each discovered from the previous one's error:

```bash
hedgehog override add TASKS-CONTRACT --scope 'packages/contracts/*' --scope 'packages/contracts/src/*' --reason '...'
# → "run hedgehog plan --recompile to widen TASKS-CONTRACT now, if it's already compiled"
hedgehog plan --recompile
# → "0 updated, 1 skipped" (task is leased to me)
hedgehog release TASKS-CONTRACT --owner builder
hedgehog plan --recompile
# → "skipped TASKS-CONTRACT blocked — pass --include-blocked to rewrite anyway"
hedgehog plan --recompile --include-blocked
hedgehog retry TASKS-CONTRACT
hedgehog claim TASKS-CONTRACT --owner builder
```

**Credit where due:** every one of those errors named its own fix precisely, which is why this was navigable at all rather than a dead end. It is the number of round-trips that is the problem, not the messaging.

## Suggested fix

The skill already knows exactly which tasks need this:

> Which tasks need it: the first module through `contract` (`packages/contracts`) and through `hook` (`packages/hooks`), and every module's `repository` and `service` …

Since that set is statically known, surface it in the packet itself — a `FIRST ARRIVAL` line next to ALLOWED SCOPE, printing the exact `override add` command for that task. `hedgehog claim` could even detect it directly: the layer creates a package that does not exist on disk yet.

I hit this twice (`contract`, then `hook`). The second time I pre-widened before scaffolding and it was completely painless — one command, no blocked state, no recompile. That is evidence the mechanism is right and only its discoverability is off.

Environment: full-stack-app core, single module, all add-ons off.

### LINKED ISSUE #125: schema and contract generators hardcode 255-char strings, silently diverging from the intent's own rules
My intent carried an explicit rule:

> Title is required: it must be present, non-empty after trimming whitespace, and **at most 500 characters**.

That rule is printed verbatim in the RELEVANT RULES block of both the `schema` and `contract` packets. Both generators nonetheless emit 255:

`packages/db/src/schema/tasks/tasks.table.ts`
```ts
title: varchar('title', { length: 255 }).notNull(),
```

`packages/contracts/src/tasks/tasks.schema.ts`
```ts
export const taskSchema = z.object({
  title: z.string().min(1).max(255),
  ...
});
export const createTaskSchema = z.object({
  title: z.string().min(1).max(255),
  ...
});
```

`--fields` has no length syntax (`name:type` with a trailing `?` for nullable), so 255 is unavoidable from the generator and must be hand-corrected in two packages, in two separate layers, several steps apart.

## Why it is worth catching

The failure is silent and split across layers. If you fix the Zod `.max()` but not the Drizzle `varchar`, a 300-char title passes validation and fails at the database with a driver error — a 500, which directly violates the "never returns a 500" rule the same intent states. Nothing in the gate compares a layer's output against the RELEVANT RULES it was handed; `verify` only runs that layer's own tests, which the generator did not write for this.

## Suggested fix

In rough order of preference:

1. Extend `--fields` with an optional length: `title:string(500)`, threaded to both `varchar(...)` and `.max(...)`. `contract` already takes "the same list the module's `schema` was generated with", so the two stay consistent by construction.
2. Failing that, have the generators emit a named constant with a visible `TODO`, rather than a plausible-looking `255` that reads as deliberate.
3. Document in `hedgehog-loop`'s "Scaffolding a layer" that string lengths are always part of the authored delta and must be checked against the packet's rules — currently that section names "the field list and its types" as the delta, which sounds like `--fields` already covered it.

Related: `drizzle-zod`'s `createInsertSchema` does derive the length bound from the column, so option 1 in `schema` alone gets most of the way there — but the `contract` layer hand-writes its own `z.string().max(255)` rather than deriving, so it would still need the same value.

### LINKED ISSUE #126: hook generator's --toggleField builds a client-side read-modify-write, reintroducing the race a server toggle prevents
Ran the documented invocation:

```bash
nx g ./tools/generators:hook --module=tasks --toggleField=completed
```

It emits:

```ts
export function useToggleTask() {
  const update = useUpdateTask();

  return {
    ...update,
    mutate: (variables: { id: string; completed: boolean }) =>
      update.mutate({ id: variables.id, completed: !variables.completed }),
    mutateAsync: (variables: { id: string; completed: boolean }) =>
      update.mutateAsync({ id: variables.id, completed: !variables.completed }),
  };
}
```

The new value is computed **on the client, from a cached row**, and PATCHed as an absolute value. Two toggles racing from a stale cache both read `completed: false` and both write `true`, so one flip is lost. It is a read-modify-write with the read served from TanStack Query's cache rather than the database.

## Why this is a notable default

The rest of the core is careful about exactly this. `service`'s generated `update` wraps read-then-write in a transaction with a comment explaining why:

> Read-then-write is two operations against one row, so it runs in one transaction: without it a concurrent delete between the two leaves `update()` returning undefined for a row this method already proved present.

`--toggleField` then hands the same hazard to the client, where there is no transaction available at all.

## What I did instead

Added a dedicated contract route whose body is `c.noBody()`:

```ts
toggle: {
  method: 'POST',
  path: '/tasks/:id/toggle',
  pathParams: z.object({ id: z.uuid() }),
  body: c.noBody(),
  responses: { 200: taskSchema, 404: taskNotFoundSchema },
},
```

with the service flipping from stored state inside the existing transaction, and a hook that sends only the id. The client never transmits a `completed` value, so a stale cache cannot overwrite a newer state.

## Suggested fix

Either make `--toggleField` generate the server-side endpoint shape (contract route + service method + controller handler + id-only hook), or drop the flag and document the pattern in `hedgehog-loop`. The current default looks correct in review — it produces a `useToggleTask` with a clean signature — which is what makes it easy to keep.

Minor related note: `@ts-rest/core` generates **no** `body` parameter for a `c.noBody()` route, so `client.toggle({ params: { id }, body: undefined })` is a TS2353 compile error. Correct call is `client.toggle({ params: { id } })`. Caught by `hedgehog verify`'s typecheck, not by the layer's own test command.

### LINKED ISSUE #127: apps/api exits with code 1 when Postgres restarts under it (unhandled pg pool error)
Reproduced on a stock full-stack-app core, all add-ons off, while verifying the "tasks persist between restarts" requirement.

## Reproduce

```bash
docker compose up -d
nx serve api
curl "http://localhost:3333/api/tasks"     # 200
docker compose restart postgres
curl "http://localhost:3333/api/tasks"     # connection refused — process is gone
```

`nx serve api` log ends with:

```
    at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
 NX  Process exited with code 1, waiting for changes to restart...
```

An idle connection in the `pg` pool is severed when the server goes away. `pg` emits `'error'` on the pool for idle-client failures, and with no listener attached Node treats it as an unhandled `'error'` event and terminates the process.

## Impact

Data is unaffected — this is purely process resilience. I confirmed a cold start reconnects cleanly with every row intact, including completion state and due dates, after restarting **both** the API process and the Postgres container.

But it means any transient database blip (container restart, failover, laptop sleep, `docker compose restart`) kills the API rather than reconnecting. In dev, `nx serve`'s watcher does not restart it either — it sits at "waiting for changes" until you touch a file. In deployment (Railway, per the locked stack) it would be a crash-loop on every routine Postgres restart.

## Suggested fix

Attach a pool error handler in `packages/db`'s client — roughly:

```ts
pool.on('error', (err) => {
  logger.error({ err }, 'idle postgres client error');
  // do not rethrow: pg discards the bad client and the pool creates a new one
});
```

`pg` already removes the failed client from the pool, so swallowing it (with a log) restores the intended behaviour: the next query opens a fresh connection. Worth pairing with explicit `connectionTimeoutMillis` / `idleTimeoutMillis` so a wedged connection surfaces as a query error rather than a hang.

Since this lives in the shipped `src/golden-cores/full-stack-app/` tree, it is inherited by every project on this core rather than being a per-project fix — which is why I am filing it rather than patching locally. I recorded it as inherited debt on the controller task in my own build, but the fix belongs upstream.

Environment: macOS (Darwin 25.3.0), Docker Desktop, postgres:16-alpine, Nx 23.1.0, all add-ons off.

### LINKED ISSUE #131: Core ships a root page with a dead "Get started" button that no layer is scoped to fix
The full-stack-app core lands `apps/web/src/app/page.tsx`:

```tsx
export default function Index() {
  return (
    <div className="flex min-h-screen flex-col items-center justify-center gap-6 p-8">
      <div className="flex items-center gap-4">
        <h1 className="text-2xl font-semibold">Hedgehog</h1>
        <ThemeToggle />
      </div>
      <Button>Get started</Button>
    </div>
  );
}
```

The `Button` has no `onClick` and no `Link` — it is a ShadCN smoke test, which is a reasonable thing for the shipped core to contain. The problem is what happens after a build finishes.

## No layer can fix it

`core.yaml`'s only `apps/web` scope is the screen layer's:

```yaml
scope: ["apps/web/src/app/{module}/**"]
```

That is module-scoped by construction (`validateCore` requires `{module}` on a module-axis core), so `apps/web/src/app/page.tsx` — one level up — is outside **every** layer's ALLOWED SCOPE. A screen layer that edited it would fail `hedgehog verify` exactly the way a first-arrival package shell does, and the packet's own HONESTY block says a scope that seems wrong is "reported, not widened."

So the correct behaviour, which I followed, is to build the module screen at `/{module}` and leave the root page untouched. The result is that a completed build ships a landing page whose only affordance is a button that does nothing, with no route to the app that was just built. There is no post-`join` step that owns this, and the module axis cannot reach it.

## Why it reads as a defect to a user

I built a tasks module at `/tasks`. Visiting `localhost:3000` gives "Hedgehog" and a dead "Get started". Nothing links the two. A user landing on the root of their own finished app sees a placeholder and no way in — the app appears broken even though every task in the graph is `complete` and the INTENT CHECK passed.

## Suggested fix

Options, roughly in order of how much they respect the existing design:

1. **Make the shipped root page self-describing rather than actionable.** Drop the `Button` entirely; render a short "no modules built yet" state. A page that promises nothing cannot be a dead end. This is the smallest change and needs no scope work.
2. **Give the root page a generated route index**, the way `apps/api` already solves the identical problem: `tools/generate-feature-modules.cjs` globs `apps/api/src/app/*/*.module.ts` so no module ever edits `app.module.ts`. The same trick would work here — glob `apps/web/src/app/*/page.tsx` into a generated list the root page renders, keeping it out of every module's scope while staying current. This seems most in keeping with the core's own conventions.
3. **Add a project-wide finishing layer** (`exclusive: true`, like `join`) that owns app-shell concerns — root page, nav, layout. Heavier, but it gives the shell an owner instead of leaving it ownerless.

Option 2 is appealing because the precedent already exists in this codebase and the reasoning is written down in `hedgehog-bootstrap-full-stack-app-core` ("The controller barrel"): a shared file that no module-scoped task can safely touch gets generated, not hand-edited. `page.tsx` is the same category of file, on the web side, without the same solution.

Environment: full-stack-app core, one module (`tasks`), all add-ons off, build completed 8/8.

skyf0xx:
Works the inbound bug queue: eight open issues, each fixed at the layer that owns the defect rather than documented around.

Closes #121
Closes #122
Closes #123
Closes #124
Closes #125
Closes #126
Closes #127
Closes #131

## Runtime and build-graph fixes

**#127 — `apps/api` exited with code 1 when Postgres restarted under it.** `pg` emits `'error'` on the pool when an *idle* client's connection dies (container restart, failover, laptop sleep), and Node terminates the process on an unhandled `'error'` event — so a routine database blip crash-loops the API instead of reconnecting. `pg` has already discarded the bad client by then, so logging and swallowing is the recovery. Paired with explicit `connectionTimeoutMillis`/`idleTimeoutMillis` so a wedged connection surfaces as a query error rather than a hang.

**#123 — `hedgehog db rebuild` left stale task rows behind.** Replay was additive, so an intent file that was renamed or deleted left its tasks in the graph; the scheduler then correctly saw a ghost task whose scope overlapped the real one and held the real one back, producing a graph no set of committed intents describes. The derived graph is now cleared before replay, making the result the pure function of the committed sources the docs already claim. `debt` and `friction` are the deliberate exception — operator-recorded, with no committed source to replay from — so they are carried across by task id and re-attached; notes whose task the recompile no longer holds are printed with their text rather than disappearing into a count.

**#122 — a singular intent id silently decided every scope glob.** The id is chosen at planning intake, several steps before anything states the module axis is expected to be plural, and nothing caught it: `intent add` accepted it, `plan` compiled it, `claim` handed out a packet whose own `scaffold from:` line contradicted its `ALLOWED SCOPE`. Now warns at `intent add`, where the fix is still one file rename, and states the convention in `hedgehog-planning-intake` where the id is actually chosen rather than only where it is consumed. A warning rather than a refusal — plenty of real modules are singular (`billing`, `search`), so the convention can't be enforced without rejecting correct ids. Also extracts `isModuleAxis` in `core.mjs` as the single owner of that question, which `validateCore` and `lintCore` each computed inline.

**#124 — first-arrival widening was documented as pre-emptive but only discoverable after a failed verify.** By the time `verify` rejected the package-shell paths the task was `blocked`, and "widen it before building" was no longer available — a five-command recovery. The signal is on disk, so the packet now says so: a `FIRST ARRIVAL` section under `ALLOWED SCOPE` carrying the exact `override add` command for that task. The package root resolves with `{module}` substituted first, so a root *below* the module segment (`libs/<module>/repository`) works as well as one above it (`packages/contracts`).

## Generator fixes

**#125 — `schema` and `contract` hardcoded 255-char strings.** The intent's own rule ("at most 500 characters") was printed verbatim in both packets' `RELEVANT RULES` and both generators emitted 255 anyway. The failure is silent and split across layers: fix the Zod `.max()` but not the Drizzle `varchar` and a 300-char title passes validation to fail at the driver as a 500 — violating the same intent's "never returns a 500" rule. Adds an optional length to the field DSL (`title:string(500)`), threaded from the one parsed field to both the `varchar` and the `.max()` so the two bounds cannot disagree. A length on a type that has none is rejected rather than silently dropped; omitted, it stays 255.

**#126 — `--toggleField` built a client-side read-modify-write.** The new value was computed on the client from a cached row and PATCHed as an absolute value, so two toggles racing from one stale cache both wrote the same value and lost a flip. The rest of the core is careful about exactly this — `service`'s generated `update` wraps read-then-write in a transaction with a comment explaining why — and this handed the same hazard to the client, where no transaction is available at all. `--toggleField` now threads through `contract`, `service`, `controller` and `hook`, so one flag generates the route, the domain method, the handler and the mutation from one source: `POST /<module>/:id/toggle` carries `c.noBody()`, and the service flips from stored state inside the same transaction its `update` uses. The client sends only the id, so a stale cached row cannot overwrite a newer state.

Also fixes a pre-existing bug this exposed: the controller generator emitted an unused `toDate` helper for any module with no timestamp field, failing that layer's own typecheck.

**#131 — the core shipped a root page with a dead "Get started" button.** `core.yaml`'s only `apps/web` scope is the screen layer's, module-scoped by construction, so `apps/web/src/app/page.tsx` one level up sat outside every layer's `ALLOWED SCOPE` and no layer could fix it — a completed build shipped a landing page with no route into the app it had just built. Applies the answer `apps/api` already uses for the identical shape (a shared file no module-scoped task can safely touch is generated, not hand-edited): `tools/generate-module-routes.cjs` globs `apps/web/src/app/*/page.tsx` into `module-routes.ts`, which the root page renders as one link per module, or a short "no modules yet" line when the array is empty — which is what the shipped core has.

## Docs

**#121 — newly-installed agents are not registered in the session that ran `init`.** A host enumerates its agents once, at session start, and `init` writes them after that. The worse symptom has none: names like `planner` and `reviewer` often resolve to the user's own unrelated *global* agents, silently running a different discipline rather than erroring. `init`'s next step now says to start a new session and why, and the project `CLAUDE.md` names both symptoms so neither is mistaken for a broken install, with reading the agent file and following it inline as the in-session fallback.

## Verification

`pnpm check` and all 56 reproductions pass. Two new repros cover the fixes with the subtlest failure modes: `rebuild-replaces-graph` (ghost tasks, plus note carry-across and the orphan report) and `first-arrival-in-packet` (both package-root shapes, and the section disappearing once the package exists).

The generator changes were verified against a real workspace rather than by inspection — a full `widgets` module generated through all six layers, confirming `varchar(500)` and `.max(500)` agree, the toggle is server-side, and `typecheck` + `test` pass across all six projects; the scratch output was then reverted. The route index was verified in both states (empty and populated), with `nx run web:build` picking up `generate-module-routes` as a dependency.

## Versioning

Bumps `package.json` to **4.3.3** — the changes here are under `src/` and `bin/`. Per `CLAUDE.md`, the tag is deliberately **not** pushed: `publish.yml` creates `v4.3.3` itself against the merge commit, and a hand-pushed tag collides with its own `git push origin "$TAG"`.

The plugin version (`.claude-plugin/`, `.cursor-plugin/`) was already bumped by the earlier `feat: improve plugin` commit on this branch, which is the payload — `skills/` and `hooks/` — that version tracks. The two move independently and neither was touched by the other's changes.


skyf0xx:
Note on the one red check — **PR Risk Labelling** — which is a pre-existing CI bug, not a problem with this diff.

The job scans the diff and, on a match, writes `match=true` and then emits notices with:

```bash
[ "$GOT" -lt "$CHANGED" ] && echo "::notice::..."
[ -n "$BLIND" ] && echo "::notice::..."
```

The step runs under `bash -e`. When those tests are false the line returns 1, and as the last command in the branch that fails the whole step — immediately after `match=true` was written. The labelling step is gated on that output, so it gets skipped: the job fails loudly *and* the `risk:execution` label it had just decided to apply never gets applied. That is exactly what happened here (every other risk label is on this PR; `risk:execution` is not).

Fixed in 9e30663 by making both notices `if` blocks, which carry no such exit-status edge. Scan logic and fail-safe behaviour are unchanged. Verified locally against the old and new code: old exits 1 after writing `match=true`, new exits 0, and all three fail-safe paths (truncated file list, blind/oversized patch, clean no-match) emit their notices and exit 0.

The check stays red on this PR regardless, because `pr-risk.yml` runs on `pull_request_target` — GitHub executes the workflow file from the **base** branch, so this run is still using `master`'s buggy copy rather than the fixed one in this PR. That is the intended security property (it is why the job reads the diff as data and never checks out PR code), not something to work around. The fix takes effect for every subsequent PR once this merges.

Everything that gates correctness — Check, CodeQL, PR Policy, Dependency Review — is green.

Issue #140, opened 2026-08-14
Self update plugin


skyf0xx:
### What problem does this solve?

Stale plugins

### Proposed change

On invoke, the plug-in should run an npx update before building

PR #141, opened 2026-08-14
feat: detect and report stale Hedgehog installs

skyf0xx:
Hedgehog ships close to daily, so an installed copy goes stale fast and
nothing tells anyone. Two things go stale independently, and they update
by different mechanisms:

| | What goes stale | How it updates | Who owns it |
|---|---|---|---|
| **npm payload** | the agents/skills copied into a project | `npx @skyf0xx/hedgehog@latest update` | the project |
| **plugin** | the global offer gate + hooks | `claude plugin update hedgehog` | the host |

A user can easily have a current plugin and a year-old repo payload.

## Detecting a stale payload

`init` and `update` stamp the version they wrote into
`.hedgehog/version.json`. Nothing recorded this before, so an installed
payload was indistinguishable from any other release's.

That stamp is compared against the newest published release two ways:

- **Passive** — a one-line notice on `status` and `next`, so a stale
  project surfaces itself without anyone thinking to ask.
- **Explicit** — `hedgehog update --check`, which exits 1 when a newer
  release exists and writes nothing.

The lookup is cached for a day and best-effort: offline, timeout, or an
unstamped project all read as "no news". The notice goes to stderr and is
skipped under `HEDGEHOG_NO_UPDATE_CHECK`, so it can't corrupt anything
parsing stdout.

## Every host, not just Claude Code

The notice lives in the CLI rather than in any host's hook, because all
three hosts drive the same binary — 12 skill/agent files call `hedgehog
status`/`next`, and those ship verbatim to every host. One implementation
reaches Claude Code, Cursor, and Gemini CLI, including Gemini CLI, which
has no hook mechanism at all.

Verified on a real three-host install: one `update` refreshed 165 files
across all three in a single pass.

| Host | Payload | Plugin |
|---|---|---|
| Claude Code | CLI notice + `--check` | session-start notice |
| Cursor | CLI notice + `--check` | session-start notice |
| Gemini CLI | CLI notice + `--check` | n/a — no plugin exists |

The plugin check is Claude Code and Cursor only by design, not oversight:
Gemini CLI installs the npm payload directly and has no plugin to be
stale.

## Detecting a stale plugin

The host owns the plugin directory, so the `SessionStart` hook reports
and names the command rather than writing behind the host's back. The
notice is independent of the `.hedgehog` offer gate, making it the one
thing emitted into an already-installed project.

## Two bugs found while building this

Both would have made the feature silently never fire:

- **Registry 406.** npm's abbreviated-metadata `accept` header is only
  valid on the packument root; sending it to `/latest` returns 406, which
  the `res.ok` guard swallowed as "no answer". Dropping the header fixes
  it and returns ~47KB less data.
- **Reading the working tree.** The plugin check first read the
  marketplace clone's checked-out manifest — which is whatever the host
  last pulled, so a published update is invisible there until the host
  refreshes, precisely the staleness being checked for. This repo's own
  machine proved it: installed 1.2.0, working tree 1.2.0, `origin/master`
  1.3.0. It now reads `origin/<branch>` via `git show`, with a detached
  fetch keeping that ref current.

## Verification

`npm run check` and all 58 reproductions pass. Beyond that, tested
against the live registry (stale/current/ahead), the real stale 1.2.0
plugin, and the failure paths that must never break a session start — no
`CLAUDE_PLUGIN_ROOT`, missing clone, `version: "unknown"`, local ahead of
marketplace, and Cursor's `additional_context` field. All emit valid
JSON. Hook runtime ~65ms with no network wait; cache confirmed at 398ms
first call, ~77ms after.

## Versions

npm 4.3.6 → 4.4.0 (`src/`, `bin/`) and plugin 1.3.0 → 1.4.0 (`hooks/`),
bumped independently per their own triggers. The plugin bump is what a
marketplace update check reads to know a new version exists — the
mechanism this change depends on.

No tag pushed: `publish.yml` creates `v4.4.0` from the merge commit.

## Known limit

The plugin check detects staleness one session late on a cold clone: the
first session reports against the already-fetched ref, and the fetch it
starts lands for the next. Blocking session start on a network call
seemed the worse trade.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "hedgehog",
       "description": "Offers to set up the Hedgehog build discipline in any project you open",
-      "version": "1.3.0",
+      "version": "1.4.0",
       "source": "./",
       "skills": [
         "./skills/"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "hedgehog",
   "description": "Offers to set up the Hedgehog build discipline in any project you open",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "author": {
     "name": "skyf0xx"
   },
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "hedgehog",
   "description": "Offers to set up the Hedgehog build discipline in any project you open",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "author": {
     "name": "skyf0xx"
   },
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -132,16 +132,23 @@ discipline's stance and rationale.
   when its format differs from the canonical one. `capabilities.mjs`
   owns the agent → tool-grant fact for hosts that can't register a
   per-agent grant; `routing.mjs` generates the root `AGENTS.md` index
-  from the agents' and skills' own frontmatter. See
+  from the agents' and skills' own frontmatter. `version.mjs` owns
+  payload-staleness detection: `init` and `update` stamp the version they
+  wrote into `.hedgehog/version.json`, and that stamp is compared against
+  the newest published release. See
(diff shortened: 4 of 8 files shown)


skyf0xx:
Resolves https://github.com/skyf0xx/hedgehog/issues/140

Issue #145, opened 2026-08-15
Feat: Refactor to enable growth of golden cores


skyf0xx:
Why
The repo os getting popular and it's likely more opinionated cores and languages will be added soon.

What
To allow for this, we need to refactor Hedgehog so that:
1. Golden cores are separate, easy to maintain and extend repos with agents, skills, generators, etc
2. Hedgehog has a registry of repos it can use

skyf0xx:
completed https://github.com/skyf0xx/hedgehog/pull/146

PR #146, opened 2026-08-15
BREAKING: v5.0.0 — cores install from their own npm packages

skyf0xx:
## Breaking change

Cores install from their own npm packages
(`@skyf0xx/hedgehog-core-full-stack-app`, `@skyf0xx/hedgehog-core-landing-page`,
`@skyf0xx/hedgehog-core-authored`), fetched by name at install time rather than
shipped inside this package. A project installed on a 4.x engine has no recorded
core, and `update` refuses to touch it rather than guessing which core to refresh
or silently deleting its agents and skills. There is no automatic migration. The
path forward is reinstalling with `init --core <name> --force`, naming the core
the project actually uses.

## What this is

The engine (`@skyf0xx/hedgehog`) keeps the methodology, CLI, build graph, host
adapters, and a registry naming the cores it knows about. `init` resolves a core
by name from `src/registry/cores.json`, fetches it from npm, and caches the
extraction locally so a repeat install needs no network. Each core ships as its
own versioned package holding that core's workspace, agents, skills, and
CLAUDE.md section.

All three core packages are published and resolvable at `1.0.0` on the public
npm registry, verified end to end against real registry resolution: fresh
installs of all three cores, a second host added to an existing project, `update`
on a normal project, `update`'s refusal on a project with no recorded core, and
offline behavior with a warm cache — all confirmed with real command output, no
local-source override.

This PR also closes four gaps found during that verification: the engine version
now matches the `^5.0.0` every core manifest requires; a core whose manifest
declares an engine range this CLI doesn't satisfy is refused with a clear message
naming both versions; the core-selection prose (`selects_when`) has a single
owner (the registry, which is what `planner` reads before any package is
fetched) instead of being duplicated into each core's manifest; and a local
fixture-drift check catches a core's shipped manifest diverging from the copy
`scripts/check.mjs` validates against.

## Verification

```
npm run check   → ok — 4 agents, 4 skills, 3 cores, tarball, CLI
npm run repro   → all 56 reproduction(s) passed
npm pack --dry-run → version: 5.0.0, src/registry/cores.json present, no src/golden-cores/
```

## Not in this PR

The engine is not published to npm at v5 by this PR — merging to `master` is
what triggers `.github/workflows/publish.yml` to tag, publish, and create the
GitHub release. Nothing here publishes or tags by hand.


skyf0xx:
resolves https://github.com/skyf0xx/hedgehog/issues/145

Issue #298, opened 2026-08-26
Global CLI install has no per-project version pin, so upgrading it anywhere breaks every other project's committed DB


skyf0xx:
**What happened**

`@skyf0xx/hedgehog` is installed once, globally (`npm install -g @skyf0xx/hedgehog`), and shared across every Hedgehog project on the machine. On 2026-08-26 I ran `npm install --global @skyf0xx/hedgehog@6.0.4` from inside the Hedgehog tool's own source checkout (`~/Documents/code/hedgehog`), which replaced the one global binary everywhere — including for an unrelated browser-extension project (`~/Documents/code/browser`) whose `.hedgehog/hedgehog.db` had last been written on 2026-08-03 by whatever CLI version was current then (6.0.2 or 6.0.3, judging by the registry's version list).

The next `hedgehog status`/`hedgehog next` call in that unrelated project failed outright:

```
Install failed: no such column: t.lease_owner
```

There's no `hedgehog db migrate` (or equivalent) command — `hedgehog --help` has nothing migration-shaped, and the only DB-mutating subcommand I found (`hedgehog db rebuild`, per #123) replaces the graph rather than migrating it in place, which isn't safe to reach for on a project with real in-progress build state.

**Why this matters**

`.hedgehog/hedgehog.db` is committed to git and documented as the project's live source of truth ("never re-derive build state from prose" — CLAUDE.md). But the CLI reading/writing it is a *global*, unversioned-per-project install. That means:

- Upgrading the CLI for **any reason, in any project** (including working on Hedgehog itself) can silently brick every *other* Hedgehog project on the same machine whose committed DB predates the new schema.
- The break surfaces as a raw SQLite error (`no such column`) with zero indication that a version mismatch, rather than DB corruption, is the cause.
- There's no recovery path that preserves existing graph state — only `db rebuild`, which starts over.
- This is worse on a team: whoever happens to have the newest global CLI installed (for unrelated reasons) can push a DB-incompatible commit without realizing it, breaking `hedgehog status` for every teammate who pulls it, or vice versa.

**Suggested fixes** (not mutually exclusive)

1. **Record the schema/CLI version the DB was written with**, either as a row in the DB or a sibling `.hedgehog/hedgehog.lock`, and have every command check it before touching the DB — fail with a clear "this DB was written by 6.0.3, you're running 6.0.4; run `hedgehog db migrate`" instead of a raw SQL error.
[...]

PR #307, opened 2026-08-27
feat(db): versioned schema migrations and a decisions ledger

### LINKED ISSUE #301: Add schema versioning and forward migrations to prevent CLI upgrade breakage
See https://github.com/skyf0xx/hedgehog/issues/298#issuecomment-5434094889

Need to add:
1. Schema version tracking — record the CLI/schema version that wrote the DB, show clear errors on mismatch instead of raw SQL
2. `hedgehog db migrate` command for forward migrations so upgrades don't force a graph rebuild

Both in one PR. Preserving graph state across CLI bumps is worth doing right rather than leaving users with just "rebuild and lose your history."

Bryandero98:
Fixes #301.

Combines #301's literal ask (versioned schema migrations) with the Decision Ledger proposed on #298 — both build on the same `applySchema`/`openDb` self-healing path from #299.

## Schema Versioning

- `PRAGMA user_version` tracks schema version; `CURRENT_SCHEMA_VERSION` + an ordered `MIGRATIONS` array (`src/db/schema.mjs`) replace the single `ensureTaskColumns` call.
- `hedgehog db migrate` reports the version transition or "already at the latest schema."
- Opening a graph with `user_version` ahead of `CURRENT_SCHEMA_VERSION` throws instead of failing on an unrecognized column later.

## Decisions Ledger

- New `decisions` table (`src/db/decision.mjs`), `hedgehog decision add/list`, and an `INHERITED DECISIONS` packet section (`src/db/next.mjs`) — same delivery path as `debt`, for *why* a task was built a certain way rather than what's wrong with it.
- `hedgehog db rebuild` carries `decisions` across a rebuild by task id (`src/db/rebuild.mjs`), same as `debt`/`friction`.

## Fixes

- Corrected `debt.mjs`'s comment claiming debt doesn't survive `db rebuild` — it does (verified by reading `rebuild.mjs`'s carry-across logic).
- `repro/packet-sections-preserved.mjs` hardcoded the packet section list without `INHERITED DECISIONS`; updated its baseline and expected order (flagged in review).

## Test plan

- Ran `node --experimental-sqlite repro/schema-migration-version-tracking.mjs` — pass (v0→v2 migration, no-op re-run, newer-than-CLI rejection).
- Ran `node --experimental-sqlite repro/decision-reaches-inheriting-packet.mjs` — pass (decision inherited into `next`/`claim` packets, `decision list` reports it).
- Ran `node --experimental-sqlite repro/run-all.mjs` (62 scripts) — same pre-existing unrelated failures as `master`, none newly introduced.


skyf0xx (changes requested):
The `check` CI job is failing: `repro/packet-sections-preserved.mjs` hardcodes the packet section list and doesn't include `INHERITED DECISIONS`, so it fails now that this PR adds that section. 

Can you update that repro's expected output to include the new section? @Bryandero98 

Everything else looks solid.

Bryandero98:
Fixed and pushed - `packet-sections-preserved.mjs` now includes `INHERITED DECISIONS` in its baseline and expected order. CI is green (`check`, `CodeQL`, `analyze` all passing on 6922f8e).

skyf0xx (approved):

skyf0xx:
@Bryandero98 nice work! - runs as expected and recording the debt really should help as well

Issue #313, opened 2026-08-27
Improve how Hedgehog tracks and enforces a project's architecture


skyf0xx:
## Why

1. Every project Hedgehog builds already has an architecture, but that architecture is not explicitly named and tracked.
2. On an existing codebase, Hedgehog silently copies whatever structure is already there, good or bad, with no way to flag it.
 
To fix this:
- We want every project, new or existing, to have its architecture named and automatically checked.
- We want existing codebases to get a heads-up when new work has to bend around a structural problem and either continue or refactor around the problem.

## What

Seven sub-issues, in dependency order. #314 is the foundation everything else needs; #319 and #320 are unrelated to the `pattern` field and can be picked up any time.

1. #314 — add and enforce the `pattern` field (engine) — **start here**
2. #315 — declare `pattern` on the four shipped cores
3. #316 — blueprints declare `pattern`; `hedgehog-core-design` carries it through
4. #317 — `hedgehog-adopt` observes and records a repo's existing pattern
5. #318 — opt-in `pattern_target` + mechanical debt note (gated — not open yet, see issue)
6. #319 — test depth follows verify radius, review follows exclusivity
7. #320 — state `hedgehog-adopt`'s change-work sizing bar as conditions, not examples

<details>
<summary>Full reasoning (for AI agents and anyone who wants the detail)</summary>

## The observation this rests on

**The layer chain is already an architecture commitment. Nothing declares which one.**

- `blueprints/cli.md` (in `@skyf0xx/hedgehog-core-authored`) derives `command → domain → io-adapter` and states "never let `command` reach past `domain` into `io-adapter`." That is ports-and-adapters, asserted as an unnamed one-off boundary rule.
- `full-stack-app`'s `schema → contract → repository → service → controller`, instantiated per `{module}`, is a layered shape sliced vertically by domain module.
- `landing-page` and every adopted repo's `hedgehog-adopt` chain are a strict linear chain.

Three architectures ship today, hardcoded per blueprint, none of them named, and none of them checked. The dependency rule everyone writes prose about — "the domain must not depend on infrastructure" — is not prose here: it is a layer's scope globs not containing adapter paths, and a layer having no `depends_on` edge to an adapter layer. **The architecture is the graph.** Declaring it makes it choosable deliberately and checkable mechanically.

## Why a field and not a skill

[...]

PR #325, opened 2026-08-28
feat(core): parse and conformance-check a `pattern` field

### LINKED ISSUE #314: core.yaml: parse and conformance-check a `pattern` field
Part of #313. **Build first — every other issue in that series depends on this one.**

## Goal

Add an optional top-level `pattern` key to `core.yaml`, and make `validateCore` check that a core claiming a pattern actually has that pattern's dependency shape. Engine-only: `src/db/core.mjs`, plus a repro and the `hedgehog status` surface.

## Why the check is the point

An unchecked field is a comment, and a comment that can silently disagree with the graph is worse than no field at all. The value here is that the dependency rule every architecture doc writes prose about — "the domain must not depend on infrastructure" — is already expressible in data Hedgehog parses: a layer's `depends_on` edge and its `scope` globs. The check reads what is already there.

## The field

```yaml
id: my-core
pattern: hexagonal     # optional; one of hexagonal | layered | vertical-slice | none
layers:
  - id: domain
    ...
```

Absent means unset, and an unset pattern is **not** an error — every `core.yaml` written before this field exists must stay valid, the same stance `requires` and `pluralizes` already take (`src/db/core.mjs:546-559`, `:218-228`). An unset pattern skips conformance checking entirely.

### 1. Parse it — `parseCoreYaml`, `src/db/core.mjs:196`

The top-level loop (`:210-228`) matches `id` and `pluralizes` by name and **silently ignores every other key**. So `pattern` needs an explicit branch or it parses to nothing:

```js
if (key === 'pattern') core.pattern = parseScalar(value);
```

Initialize `core` (`:205`) with `pattern: null` so the shape is uniform whether the key is present or not. Extend the subset comment at `:184-195` with the new key, matching the existing lines' form.

Reject an unrecognized value at parse time with the valid set named in the message — a typo'd `pattern: hexagnoal` must not silently degrade to "unset", which would turn the conformance check off without telling anyone. This is the one new throw in the parser.

### 2. Conformance-check it — `validateCore`, `src/db/core.mjs:505`

Add one block after the existing `depends_on` resolution loop (`:569-576`), so it can rely on every `depends_on` naming a real layer. Per value:

- **`none`** — no check. Explicitly valid, explicitly unchecked. This is the adopted-repo default (#D) and must never produce a finding.

- **`layered`** — every layer except the first has a `depends_on`, and the chain is linear: no two layers share a `depends_on` parent, and no layer is unreachable from the head. Throw naming the first layer that breaks it.

- **`vertical-slice`** — `isModuleAxis(core)` (`:497`) must be true. A core declaring `vertical-slice` with no `{module}` anywhere is a mislabel, and the message should say so directly. Note the existing module-axis uniformity rule (`:608-630`) already does the per-layer half of this; the new check is only the core-level claim.

- **`hexagonal`** — the real one. A core declares which layer is the domain by convention: the layer with **no `depends_on`** and no `{module}` — i.e. the head. The rule to enforce is that the domain layer's `depends_on` chain never reaches a layer whose scope is an adapter boundary. Since Hedgehog has no adapter marker today, define the check on direction alone: **the head layer must have no `depends_on`, and no layer may `depends_on` a layer that itself transitively depends on the head's dependents** — in practice, that dependencies point one way and the head is the sink. If that proves too weak to catch anything real when run against the blueprints in #C, escalate to `lintCore` rather than inventing a marker field in this PR; the marker is a separate design decision and should not be smuggled in here.

Every message follows the existing style in this file: state what was found, what the rule is, and why it matters mechanically (see `:539-542` and `:592-596` as the models to match).

### 3. Surface it — `hedgehog status`

`coreWarningLines` (`bin/cli.mjs:2511`) already prints core lint warnings. Certainties throw from `validateCore` on load and are already reported there (`:2519-2521`). Nothing new is needed for errors. Optionally print the declared pattern on the core line in `status` so a user can see what their core claims — keep it to one word, no new section.

## Tests

`repro/` is the harness (`node --experimental-sqlite repro/run-all.mjs`; the runner picks up any non-helper `.mjs` in that directory, `repro/run-all.mjs:26`). Add:

- `pattern-unset-still-valid.mjs` — a `core.yaml` with no `pattern` loads and validates exactly as today. This is the backward-compat guard and the most important one.
- `pattern-rejects-unknown-value.mjs` — `pattern: hexagnoal` throws at parse, and the message names the valid set.
- `pattern-vertical-slice-requires-module-axis.mjs` — declaring `vertical-slice` with no `{module}` throws.
- `pattern-layered-rejects-branching-chain.mjs` — two layers sharing a `depends_on` parent under `layered` throws.
- `pattern-none-never-throws.mjs` — `pattern: none` on a deliberately messy chain validates clean.

`repro/parse-real-cores.mjs` already parses the shipped cores' real `core.yaml` files; confirm it still passes untouched (it will — they have no `pattern` until #B).

## Done when

- `pattern` parses, an unknown value throws, an absent value is valid and unchecked.
- `validateCore` enforces the three checkable values; `none` is a no-op.
- Five repros above pass, and `npm run check` is green.
- `src/db/core.mjs`'s header comment and the `parseCoreYaml` subset comment both name the new key.

## Not in this issue

Declaring `pattern` on any actual core (#B, #C, #D), and anything to do with `pattern_target` (#E). This PR adds the mechanism and ships zero cores using it.

Bryandero98:
Part of #313. Closes #314.

- Adds an optional top-level `pattern` key to `core.yaml` (`hexagonal | layered | vertical-slice | none`), parsed in `parseCoreYaml` (`src/db/core.mjs`). An unrecognized value throws at parse time naming the valid set; absent stays `null` and unchecked, same stance `requires`/`pluralizes` already take.
- Adds a conformance check in `validateCore` for the three checkable values. `none` is an explicit no-op. `layered` requires a strict linear chain (every layer but the head has exactly one `depends_on`, no shared parent, every layer reachable from the head). `vertical-slice` requires `isModuleAxis(core)`. `hexagonal` checks direction alone — the head has no `depends_on`, every other layer's chain terminates at the head with no cycle — deliberately weaker than a real adapter-boundary check, per the issue's "Not in this issue" note.
- No change needed to `hedgehog status`: `validateCore` throwing on load is already caught and reported by `coreWarningLines` (`bin/cli.mjs`).

<details>
<summary>Why hexagonal is checked on direction alone</summary>

Hedgehog has no adapter marker today, so there's no way to distinguish "an adapter depending on another adapter" from "an adapter depending on the domain." The issue explicitly scopes a real marker field out of #314 as a separate design decision, so this only enforces that every layer's dependency chain points inward and terminates at the domain layer, with no cycle — real signal, not the full rule.

</details>

## Test plan

- 6 repro scripts under `repro/`, run individually and via `node --experimental-sqlite repro/run-all.mjs`: `pattern-unset-still-valid.mjs`, `pattern-rejects-unknown-value.mjs`, `pattern-vertical-slice-requires-module-axis.mjs`, `pattern-layered-rejects-branching-chain.mjs`, `pattern-none-never-throws.mjs` (the five #314 asks for), plus `pattern-hexagonal-checks-direction.mjs` (not required by the issue, added because the check is real code with real failure modes). All 6 pass.
- `repro/parse-real-cores.mjs`: confirmed via `git stash`/diff that the only change in its output is the new `"pattern": null` field on every shipped core — no other drift.
- Full repro suite (68 scripts, including the 6 new ones): ran on this branch and, separately, on unmodified `master` in the same environment. Same 10 pre-existing failures on both — all Windows-only environment gaps unrelated to this change (no `/bin/sh`, no unprivileged symlinks, Node's ESM loader rejecting a bare Windows drive-letter path, temp-dir cleanup permission errors). Zero regressions; every new pass is additive.
- `npm run check`: two of its checks fail locally on Windows, both confirmed pre-existing on unmodified `master` too, unrelated to this PR: (1) a `repro/fixtures/cores/*.manifest.yaml` fixture gets checked out with CRLF line endings on this machine, which its hand-rolled line parser doesn't strip, so every key look like it's missing (worked around locally with a `core.autocrlf=false` clone — real checkout content is LF, this is purely local `git config`, not the repo); (2) `npm pack --dry-run` fails with `spawnSync ENOENT` because Node's `execFileSync('npm', ...)` doesn't resolve `npm.cmd` on Windows. Ran `npm pack --dry-run --json` directly and confirmed `src/db/core.mjs` (the only file this PR touches) is included in the tarball. Every other check in `scripts/check.mjs` passed (it accumulates failures rather than stopping at the first one — only these two were reported). Linux CI should be green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@skyf0xx/hedgehog",
-  "version": "6.1.1",
+  "version": "6.1.2",
   "description": "Install the Hedgehog build discipline (agents + skills) into a repo, for Claude Code, Cursor, or Gemini CLI.",
   "type": "module",
   "repository": {
--- a/repro/pattern-hexagonal-checks-direction.mjs
+++ b/repro/pattern-hexagonal-checks-direction.mjs
@@ -1,0 +1,114 @@
+#!/usr/bin/env node
+// Repro: `pattern: hexagonal` checks direction alone (no adapter marker
+// exists yet — see #314's "Not in this issue"). The head layer (the
+// domain, by convention) must have no depends_on, and every other
+// layer's depends_on chain must terminate at the head with no cycle. Not
+// required by #314's own test list, but the check is real code with real
+// failure modes, so it gets the same coverage as layered/vertical-slice.
+//
+// Run: node repro/pattern-hexagonal-checks-direction.mjs
+
+import { check, finish } from './papercuts-lib.mjs';
+import { validateCore } from '../src/db/core.mjs';
+
+console.log('repro: pattern-hexagonal-checks-direction');
+
+// domain (head, no depends_on) <- repository <- controller: every chain
+// terminates at domain. Must validate clean.
+const VALID = {
+  id: 'valid-hexagonal',
+  pattern: 'hexagonal',
+  layers: [
+    { id: 'domain', scope: ['domain/**'], verify: 'true', commit: 'feat: domain' },
+    {
+      id: 'repository',
+      depends_on: 'domain',
+      scope: ['repository/**'],
+      verify: 'true',
+      commit: 'feat: repository',
+    },
+    {
+      id: 'controller',
+      depends_on: 'domain',
+      scope: ['controller/**'],
+      verify: 'true',
+      commit: 'feat: controller',
+    },
+  ],
+};
+try {
+  validateCore(VALID);
+  check('two adapters both pointing at the domain validates clean', true, {});
+} catch (err) {
+  check('two adapters both pointing at the domain validates clean', false, {
+    expected: 'no error',
+    actual: err.message,
+  });
+}
+
+// The head itself has a depends_on — it can't be the sink if it also
+// depends on something.
+const HEAD_HAS_DEPENDENCY = {
+  id: 'head-has-dependency',
+  pattern: 'hexagonal',
+  layers: [
+    {
+      id: 'domain',
+      depends_on: 'infra',
+      scope: ['domain/**'],
+      verify: 'true',
+      commit: 'feat: domain',
+    },
+    { id: 'infra', scope: ['infra/**'], verify: 'true', commit: 'feat: infra' },
+  ],
+};
+try {
+  validateCore(HEAD_HAS_DEPENDENCY);
+  check('a head layer with a depends_on throws', false, {
+    expected: 'throws',
+    actual: 'validated without error',
+  });
+} catch (err) {
+  check('a head layer with a depends_on throws', true, {});
+}
+
+// A non-head layer with no depends_on is a second, ambiguous root.
+const SECOND_ROOT = {
+  id: 'second-root',
+  pattern: 'hexagonal',
+  layers: [
+    { id: 'domain', scope: ['domain/**'], verify: 'true', commit: 'feat: domain' },
+    { id: 'orphan', scope: ['orphan/**'], verify: 'true', commit: 'feat: orphan' },
+  ],
+};
+try {
+  validateCore(SECOND_ROOT);
+  check('a non-head layer with no depends_on throws', false, {
+    expected: 'throws',
+    actual: 'validated without error',
+  });
+} catch (err) {
+  check('a non-head layer with no depends_on throws', true, {});
+}
+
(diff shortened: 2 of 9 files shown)


skyf0xx (approved):
Looks great ✓

Issue #319, opened 2026-08-27
Test depth follows verify radius; review follows exclusivity


skyf0xx:
Part of #313. Independent of the `pattern` chain (#314–#318) — can land any time.

## Goal

State, in one rule, where the real test bar and the `reviewer` pass belong, and derive it from `core.yaml`'s existing fields rather than imposing a cadence. Answers the second question raised on #305:

> TDD should only be across certain boundaries, as should be reviews - so not after each task is completed

## Current state

Half of this is already right. `reviewer` is explicitly **not** a per-commit gate:

> Use at a phase or layer transition the core's own loop skill defines... Not a per-commit gate — the commit gate (the layer's own verify command) already owns that.

And `hedgehog-authored-loop` already places it "at the last layer of a batch of change-work."

The half that is not right is testing. Every layer's `verify` runs that layer's own tests, and the loop skills state the bar uniformly:

> **Each layer's tests live inside that layer's scope** and run under its own `verify` command. A layer whose command passes with no tests is not built.

That is correct as a floor and wrong as a uniform bar. It implies the same test depth on a layer that only has to be internally consistent as on a layer that integrates everything before it — which is TDD ceremony at every gate, the "agile for agents" transformation this project exists to reject.

## The rule

`core.yaml` already carries the distinction. Two fields already mark which layers are integration points:

- **`verify_radius`** — a layer whose radius equals its `scope` has a verify command that reads only what it writes (`conflict.mjs`'s `verifyRadius()`; unset defaults to scope). A layer with a wider declared radius has a command that reads across a boundary.
- **`exclusive: true`** — the layer may not run alongside anything else. Every core's `join`-style tail layer carries it, including every adopted chain's mandatory tail.

So:

> **Test depth follows verify radius. Review follows exclusivity.**
>
> A layer whose `verify_radius` equals its `scope` is internally-consistent-only: enough tests to make its own `verify` command mean something, and no more. A layer with a wider `verify_radius`, or `exclusive: true`, is where behavior actually gets proven — that is where the real test bar and the `reviewer` pass belong.

The cadence is read off the graph's structure. Nothing new is declared, nothing is scheduled by ceremony, and a core author who has already thought about radius and exclusivity has already decided their test cadence without knowing it.

## A second consumer

[...]

PR #326, opened 2026-08-28
feat(reviewer): state the exclusivity trigger mechanically

Bryandero98:
Part of skyf0xx/hedgehog#313. Addresses skyf0xx/hedgehog#319.

Companion PRs carrying the rest of #319 (docs/skill-content only, no engine change, as the issue scopes it): the owning statement lives in `hedgehog-core-authored` — https://github.com/skyf0xx/hedgehog-core-authored/pull/10 — with `hedgehog-core-full-stack-app`, `hedgehog-core-landing-page`, `hedgehog-core-deepseek-harness`, and `hedgehog-core-adopted` referencing it by name from their own loop skills/build agents.

- `src/agents/reviewer.md`'s "When you run" named its trigger only in prose (`reviewer.md:18-21` before this PR) — a phase/layer transition each loop skill defines separately, with no shared referent. Added the mechanical fact underneath: `verify_radius` wider than `scope`, or `exclusive: true`, is a join/integration point, and that's where review belongs.
- `AUTHORING-CORES.md`'s `core.yaml` audit checklist (`AUTHORING-CORES.md:172-186` before this PR) names the same consequence at the point a new core author actually sets `verify_radius`/`exclusive` — that decision also sets test/review cadence.
- Version bumped `6.1.1` → `6.1.2` via `npm run release` (`bump-package-version.mjs` failed locally with `spawnSync npm ENOENT` — the same Windows-only `execFileSync` quirk noted on #325 — worked around by running `npm version patch --no-git-tag-version` directly and hand-updating `src/hosts/gemini/gemini-extension.json` to match, exactly what the script does). This PR carries the bump for both this branch and #325 (`feat/core-pattern-field`, open) — they were on the same base commit and I only bumped one to avoid a version-line merge conflict for whichever lands second.

## Test plan

- Read both edited files in full after editing to confirm the new text reads correctly in place and doesn't contradict anything else in either file.
- Confirmed `package.json` and `src/hosts/gemini/gemini-extension.json` agree on `6.1.2` (what `scripts/check.mjs` gates on).
- No engine code touched — no repro/test suite run needed for this PR specifically; `npm run check`'s structural checks were exercised for #325 in this same session and aren't affected by this PR's files.

### DIFF
--- a/AUTHORING-CORES.md
+++ b/AUTHORING-CORES.md
@@ -174,6 +174,12 @@ satisfies the package contract above.
 - [ ] Layer sequence matches the build order the core's own agents and
       skills actually follow.
 - [ ] Each layer's `verify` command is real and runnable, not a stub.
+- [ ] `verify_radius` (when set wider than `scope`) and `exclusive: true`
+      are declared on the layers that are actually join or integration
+      points — this is also where the real test bar and the `reviewer`
+      pass land (`hedgehog-authored-loop`'s "Test depth follows verify
+      radius"), so the decision made here is the same one that sets test
+      and review cadence, not a separate call to make later.
 - [ ] Every `verify` command has been checked for a system binary beyond
       what the workspace's package manager installs (Docker, Terraform,
       a database CLI, a compiler toolchain), and any such binary is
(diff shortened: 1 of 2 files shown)


skyf0xx (approved):
Looks really good - will rebase and merge

Issue #315, opened 2026-08-27
Declare `pattern` on the four shipped cores


skyf0xx:
Part of #313. Blocked by #314 — the field has to parse and validate before a core can declare it.

## Goal

Declare `pattern` on the four shipped cores' own `core.yaml`, so #314's conformance check has real cores to run against and every shipped core states the architecture it builds in.

## Why this is not just documentation

This is the decision recorded on #313: an unchecked field is a comment. Landing #314's check with zero cores declaring a pattern means the check never runs on anything real until a user authors a core. Declaring the four shipped cores is what proves the check works and what makes the field enforcement rather than annotation.

Each of these is a **change to that core's own package**, released from its own repo — not a change to this one. One PR per core repo.

## What each core declares

Read off the layer chain each core already ships. Nothing about any chain changes; this only names what is already true.

| Core | Package | Declares | Because |
|---|---|---|---|
| `full-stack-app` | `@skyf0xx/hedgehog-core-full-stack-app` | `vertical-slice` | `schema → contract → repository → service → controller` instantiated per `{module}` — the module axis is the slice |
| `pwa-app` | `@skyf0xx/hedgehog-core-pwa-app` | read the chain and decide | likely `vertical-slice`; confirm against its actual `workspace/core.yaml` before writing |
| `landing-page` | `@skyf0xx/hedgehog-core-landing-page` | `layered` | `brief → feeling → tokens → sequence → artifact`, a strict linear chain, no `{module}` |
| `deepseek-harness` | `@skyf0xx/hedgehog-core-deepseek-harness` | read the chain and decide | six-layer scaffold → logic → wiring → smoke → bundle → join, one plugin per intent — confirm whether the module axis is on |

**Do not force a value.** If a core's chain genuinely does not match any of `hexagonal`/`layered`/`vertical-slice`, declare `none` and say why in that PR. A wrong declaration is worse than an absent one, because #314's check will then enforce a shape the core does not have.

## Per-core PR contents

1. Add `pattern: <value>` to the core's `workspace/core.yaml` (or wherever that core keeps its shipped core definition).
2. Confirm `validateCore` accepts it — load the file through `src/db/core.mjs`'s `loadCore` against a checkout of this engine repo with #314 merged.
3. Bump that package's version per its own release process (a version bump committed and merged to that repo's `main` triggers its publish workflow).
4. One sentence in that core's own docs naming the pattern, where that core documents its layer chain.

## In this repo

[...]

PR #327, opened 2026-08-28
feat(repro): assert every shipped core declares a pattern

Bryandero98:
Part of epic skyf0xx/hedgehog#313. Addresses skyf0xx/hedgehog#315.

Companion PRs, one per core repo (all reference this issue and #313 too):
- skyf0xx/hedgehog-core-full-stack-app#23 (`pattern: vertical-slice`)
- skyf0xx/hedgehog-core-pwa-app#16 (`pattern: vertical-slice`)
- skyf0xx/hedgehog-core-landing-page#17 (`pattern: layered`)
- skyf0xx/hedgehog-core-deepseek-harness#7 (`pattern: vertical-slice`)

## What changed here

`repro/parse-real-cores.mjs` now fails if any fixture `core.yaml` omits `pattern`, so a future shipped core landing without one fails here instead of silently.

Added the missing `deepseek-harness.core.yaml` fixture - it was never vendored into `repro/fixtures/cores/` before this, so this repo's own harness had no material to test that core against at all.

Resynced all four fixtures to byte-match what each package currently ships. `pwa-app`'s fixture had drifted from reality independent of this PR: its real `workspace/core.yaml` already had two scope entries (`{module}.schema.spec.ts` on `schema`, `src/app/page.tsx` on `screen`) that this repo's fixture never picked up from that core's own earlier, unrelated work.

<details>
<summary>Two things that broke from resyncing the fixtures, both fixed alongside this (not swept under the rug)</summary>

**`repro/fixtures/shipped-cores-baseline.json` (golden snapshot) went stale.** `05-shipped-cores-unchanged.mjs` is a deliberate golden-snapshot test - its own header comment says the baseline should only be regenerated "after a deliberate edit to a shipped core.yaml, and only once the printed diff has been read line by line." That's exactly this case. Regenerated with `node --experimental-sqlite repro/05-shipped-cores-unchanged.mjs --write-baseline` and reviewed the diff: it touches exactly four `scope_globs` entries, all `pwa-app`, matching the two real scope additions above (each appears twice - once per test intent in that repro's fixture graph). Nothing else moved.

**`repro/pattern-unset-still-valid.mjs` asserted `pattern === null` on the shipped cores.** That assertion was true only because no shipped core declared a pattern yet - the whole point of #315 is to make that stop being true. Updated it to assert what that block's own name always meant ("loads and validates"), dropping the `=== null` check, and added `deepseek-harness` to the loop (it was missing, since no fixture existed for it before this PR).

</details>

No `src/` or `bin/` change, so no `package.json` version bump per this repo's own `CLAUDE.md` rule ("a change scoped to `src/` or `bin/` only bumps `package.json`"). Nothing in `src/registry/cores.json` needed widening either - all four companion PRs are patch bumps, and every one lands inside the range already pinned there (`^1.0.6`, `^1.0.4`, `^1.0.2`, `>=0.2.3 <1.0.0`).

## Test plan

Ran the full 68-script `repro/` suite (`npm run repro`) twice, on a clean `core.autocrlf=false` clone (this repo's global Windows `autocrlf=true` otherwise false-positives two unrelated checks - a YAML CRLF parse error and an `npm pack` `spawnSync ENOENT`, both already documented on #314's PR):

- **Unmodified `master`:** 11 of 68 failed, all pre-existing and Windows-only (missing `/bin/sh`, unprivileged symlinks, Node's ESM loader rejecting a Windows drive-letter path, temp-dir cleanup permissions).
- **With this change:** 10 of 68 failed - the same failures, minus one. `status-reports-drift.mjs` failed on the `master` run and passed here; reran it standalone on the unmodified `master` state (via `git stash`) and it passed clean on a second try, confirming it's flaky in this environment rather than related to this change.

No new failures introduced either way. Also individually verified: `05-shipped-cores-unchanged.mjs` and `pattern-unset-still-valid.mjs` (the two this PR's fixture resync affects) each pass clean, and `npm pack --dry-run` succeeds (run directly, since `npm run check`'s own invocation hits the Windows `spawnSync` quirk above).


skyf0xx:
Held for a human look before merging — not a rejection, everything here checks out mechanically (CI green, mergeable in any order relative to the sibling PRs since `repro/fixtures/` is vendored, not fetched from npm).

One line in the PR body doesn't hold up: "Resynced all four fixtures to byte-match what each package currently ships." I diffed the four published tarballs (full-stack-app 1.0.9, pwa-app 1.0.6, landing-page 1.0.5, deepseek-harness 0.2.6) against these fixtures — each differs from what's published by exactly the added `pattern:` line, since none of the four sibling PRs had merged yet at review time. The pwa-app scope-drift fix is genuinely a resync to what's published; the `pattern:` lines are forward-looking to what the sibling PRs add, not a resync to current reality. Worth a corrected sentence given this PR's whole purpose is fixture fidelity.

Also worth confirming intentionally rather than silently: the assertion only reaches cores with a `.core.yaml` fixture, so `authored` and `adopted` (manifest-only) aren't covered by this check — reasonable since `adopted` observes rather than declares, but worth stating if not already tracked elsewhere.

The five companion PRs (full-stack-app#23, pwa-app#16, landing-page#17, deepseek-harness#7, authored#11, adopted#3) have all merged. Happy to merge this one as soon as the fixture-fidelity claim is corrected.

skyf0xx (approved):
All this looks good @Bryandero98.

Merging

